PR #1094, opened 2014-11-05
Deprecate :member_label and :member_value options

justinfrench:
Deprecates `:member_label` and `:member_value` options in favour of simply passing a properly formatted array of labels and values into `:collection`:

```
f.input :author, :collection => [ ["Justin", 3], ["Kate", 4] ]
```

In practice, this would typically be done through a model:

```
f.input :author, :collection => Author.all.map { |a| [a.first_name, a.id] }
```

And you'd extract this out to a class method to keep the view clean, something like:

```
f.input :author, :collection => Author.names_and_values_for_collection
```

`Model.pluck` is particularly good/performant for this, allowing you to offload part of the work to the database instead of Ruby:

```
class Article
  def self.names_and_values_for_collection
    self.pluck(:first_name, :id) # ["Justin", 3]
  end
end
```

Or even combining fields into a string in SQL:

```
class Article
  def self.names_and_values_for_collection
    self.pluck(Are.sql("CONCAT(`first_name`, ' ', `last_name`)"), :id) # ["Justin French", 3]
  end
end
```

Work complete so far:
- [x] document Model.pluck in all `Collection` based inputs
- [x] mark the options as deprecated (remove in 4.0)
- [x] update DEPRECATIONS file
- [x] silence deprecation warnings in the specs

In another branch or pull-request, we can consider introducing a naming convention like `Model.names_and_values_for_collection`, which could be used as a default collection (like we do with `Model.all` currently) when it's defined on the model, simplifying the view to avoid the `:collection` option.

### DIFF
--- a/DEPRECATIONS
+++ b/DEPRECATIONS
@@ -9,6 +9,8 @@ v4.0 (planned)
 v3.1 (master)
 
 * Deprecate support for Rails < 4.1
+* Deprecate :member_value option
+* Deprecate :member_label option
 
 v3.0
 


gwincr11:
I have been moving over all my checkboxes to use the new syntax. However, when I use the new syntax in an edit mode the checkboxes that should be checked are not properly checked. Is there a new way to handle what is selected?

justinfrench:
@gwincr11 can you please open a new issue with the code you had before & after? Could be a bug, could be something simple, but it's hard to say without code :)

gwincr11:
Ok, created https://github.com/justinfrench/formtastic/issues/1125

Issue #1080, opened 2014-10-03
Deprecation on column_for_attribute with Rails 4.2.0.beta2


elia:
```
DEPRECATION WARNING: `column_for_attribute` will return a null object for non-existent columns in Rails 5.0. Use `has_attribute?` if you need to check for an attribute's existence. (called from column_for at /Users/elia/.rvm/gems/ruby-2.1.3/bundler/gems/formtastic-719a7e8098f6/lib/formtastic/helpers/input_helper.rb:300)
```

PR #1120, opened 2015-01-05
Silence deprecation we're okay with

### LINKED ISSUE #1117: Rails 4.2 deprecation on column_for
Similar to #1080, but I was advised to open a new issue. I'm using:

- Rails 4.2.0.rc3
- Formtastic 3.1.2

Formtastic is not explicitly configured, but is only used by ActiveAdmin (git HEAD, the beta of 1.0; revision f51d96dddb).

When I run my RSpec suite, I see the following deprecation warning (five or six times):

> DEPRECATION WARNING: `#column_for_attribute` will return a null object for non-existent columns in Rails 5. Use `#has_attribute?` if you need to check for an attribute's existence. (called from maxlength at /usr/local/Cellar/rbenv/0.4.0/versions/2.1.5/lib/ruby/gems/2.1.0/gems/formtastic-3.1.2/lib/formtastic/inputs/base/stringish.rb:27)

It looks to me like the trouble is at lib/formtastic/inputs/base/database.rb, line 7, where there's a call to the `column_for_attribute` method that's not wrapped in `ActiveSupport::Deprecation.silence`.

justinfrench:
Fixes #1117, similar to #1080, which was resolved in #1087 

@cpence can you try bundling this branch to ensure this works for you before I merge please?

### DIFF
--- a/lib/formtastic/inputs/base/database.rb
+++ b/lib/formtastic/inputs/base/database.rb
@@ -5,7 +5,10 @@ module Formtastic
         
         def column
           if object.respond_to?(:column_for_attribute)
-            object.column_for_attribute(method)
+            # Remove deprecation wrapper & review after Rails 5.0 ships
+            ActiveSupport::Deprecation.silence do
+              object.column_for_attribute(method)
+            end
           end
         end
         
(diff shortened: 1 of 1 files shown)


cpence:
Yep, that does it. Thanks!

PR #1124, opened 2015-01-20
testing against ruby 2.2.0

lucianosousa:


justinfrench:
@lucianosousa are you looking into the failures here?

.travis.yml
```diff
@@ -26,3 +27,7 @@ matrix:
     - rvm: 2.1
       gemfile: gemfiles/rails_edge.gemfile
       env: DEFER_GC=false RAILS_EDGE=true
+    - rvm: 2.2.0
+      gemfile: gemfiles/rails_3.2.gemfile
```
oliverklee: Does Rails 3.2 work with Ruby 2.2?

lucianosousa: still not. look at: https://github.com/rails/rails/pull/18306

oliverklee: Then we should have this as an allowed failure.

For the page_title_helper, we did it like this:
https://github.com/lwe/page_title_helper/blob/master/.travis.yml

pranas:
This is no longer relevant after #1170 was merged in. /cc @mikz

mikz:
This was fixed by #1170

PR #1126, opened 2015-02-02
Wrap checkbox and radiobutton labels with span

alesguzik:
Currently styling checkboxes and radiobuttons is only possible via hiding checkbox input and styling element  (usually `label` or `span`) next to it, as it allows to handle checkbox/radiobutton state via `:checked` pseudoclass (see http://stackoverflow.com/questions/4148499/how-to-style-checkbox-using-css).
Given formtastic places inputs inside label tags, I have implemented wrapping choice labels with `span`s.

### DIFF
--- a/lib/formtastic/inputs/base/choices.rb
+++ b/lib/formtastic/inputs/base/choices.rb
@@ -44,11 +44,13 @@ module Formtastic
         end
 
         def choice_label(choice)
-          if choice.is_a?(Array)
+          '<span>'.html_safe +
+          (if choice.is_a?(Array)
             choice.first
           else
             choice
-          end.to_s
+          end.to_s) +
+          '</span>'.html_safe
         end
 
         def choice_value(choice)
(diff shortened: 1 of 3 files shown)


mikz:
IMO this could be a good change. However, I'd rather use `content_tag` instead of `.html_safe

Fibrasek:
I really need this feature! What if I fix those conflicts, do you need something else for this feature to pass?

mikz:
@fibrasek the html attributes should be configurable the same way other elements are. And have some default options that give it some meaningful class name. 
Also would be good to ensure this is not a breaking change for the generated CSS.

justinfrench:
Closing for now due to inactivity. I'm generally supportive of this change, but…

1. the code should use `content_tag` instead of `<span>` with `html_safe`
2. I'd like to see some screenshots that prove that the default CSS does break with this change
3. It has the potential to break existing apps that aren't expecting a `<span>`. This is _okay_ for a major release, but we'll need to add a configuration option that allows older apps remove the wrapper if needed.

PR #1136, opened 2015-03-09
Remove incorrect commas in examples

maxd:
[Here](http://stackoverflow.com/questions/28943514/activeadmin-f-inputs-except-field-doesnt-work) is real problem related to these commas.


justinfrench:
Thanks!

Issue #1135, opened 2015-03-05
Setting input_class_finder fails in Rails 3.2


sepastian:
Hello!

I am working on a Rails 3.2.21 application and have updated formtastic to 3.1.3; I followed the [guide to upgrade to 3.1](https://github.com/justinfrench/formtastic/wiki/Upgrading-to-Formtastic-3.1) to get rid of the deprecation warnings.

The initializer is installed and sets `input_class_finder` and `action_class_finder`, as required.

    # config/initializers/formtastic.rb
    Formtastic::FormBuilder.input_class_finder = Formtastic::InputClassFinder
    Formtastic::FormBuilder.action_class_finder = Formtastic::ActionClassFinder

However, when I try to start the rails server, I get an error.

    ~/.rbenv/versions/2.2.0/lib/ruby/gems/2.2.0/gems/railties-3.2.21/lib/rails/railtie/configuration.rb:85:in `method_missing': undefined method `eager_load' for #<Rails::Application::Configuration:0x007f2dbc783b60> (NoMethodError)

What is the problem here?

justinfrench:
Hrm, this doesn't *feel* like a Formtastic issue given that error message, but I could be wrong.

1. What other changes were made? Can you definitively isolate this error to a tiny diff that suggests it's purely a Formtastic upgrade issue, or was there other stuff at play?
2. If you comment out the entire formtastic initializer, what happens?
3. Can you gist a full diff and a full stack trace?

zefta:
Hello,

I got the exact same error!

I got this warning
```sh
DEPRECATION WARNING: input_class is deprecated and will be removed from Formtastic 4.0 (configure Formtastic::FormBuilder.input_class_finder instead (upgrade guide on wiki: http://bit.ly/1F9QtKc )).
```

Then I ran
```sh
$ rails generate formtastic:install
```

So it added to my formatastic.rb initializer
```sh
Formtastic::FormBuilder.action_class_finder = Formtastic::ActionClassFinder
Formtastic::FormBuilder.input_class_finder = Formtastic::InputClassFinder
```

Then I got
```sh
/home/thibault/.rvm/gems/ruby-1.9.3-p551@my-project/gems/railties-3.2.21/lib/rails/railtie/configuration.rb:85:in `method_missing': undefined method `eager_load' for #<Rails::Application::Configuration:0x00000008400058> (NoMethodError)
[...]

PR #1137, opened 2015-03-10
In Rails 3, Rails.application.config does not respond to eager_load.

sepastian:
Hi, here's a PR fixing #1135.

With this patch in place, my 3.2.21 test application will boot.

However, I don't know enough about the inner workings of formtastic across different versions of Rails to assess whether or not this will break something else or not.

Can someone who knows have a look at this?

### DIFF
--- a/lib/formtastic/namespaced_class_finder.rb
+++ b/lib/formtastic/namespaced_class_finder.rb
@@ -29,7 +29,7 @@ module Formtastic
     end
 
     def self.use_const_defined?
-      defined?(Rails) && ::Rails.application && ::Rails.application.config.eager_load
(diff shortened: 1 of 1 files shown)


mikz:
So it looks like `eager_load` was introduced in `4.0.0`. This was introduced by #1100.

To maintain both current working state and rails 3 compatibility, it should do something like https://github.com/justinfrench/formtastic/blob/cdb3e81e34d0512b28761055ac7834def98870cd/lib/formtastic/helpers/input_helper.rb#L347

justinfrench:
What's the side-effect/consequences of merging this as-is @mikz?

mikz:
@justinfrench it will use the supposedly less performant method (but there is no benchmark to prove it) all the time on Rails 3. Production mode or not.

justinfrench:
@mikz okay, merge if you're happy, Rails 3 needs to be dropped anyway

Issue #1141, opened 2015-03-19
ActiveAdmin(Formtastic) and Trailblazer(Reform)


fernandes:
Hi @justinfrench,

I'm integrating AA with Trailblazer, when AA tries to render the form, show with empty fields, so I got into this

summing up: When I build a resource on AA using Trailblazer I need to pass a Reform object to the view, and AA process it and when it calls Formtastic it goes to https://github.com/justinfrench/formtastic/blob/cdb3e81e34d0512b28761055ac7834def98870cd/lib/formtastic/helpers/inputs_helper.rb#L357 and here is where I get in trouble, because:

content_columns instantiate a fresh new object from `model_name` and calls `content_columns` method that is AR specific, and worse we can't even delegate on the model inside the form object  because its instantiating a new Reform::Form object.

I was wondering how to fix this, and the best way I thought was use `@object` and call `content_columns` so we can implement on the Form object.

I'm not used to Formtastic, sorry if its a non sense question, but once you point me on the direction I can go on... and even send a PR just would like to listen to you first.

Thank you

fernandes:
@timoschilling if you could please check this issue and my proposed solutions on both PRs and suggest any better way based on AA I do really appreciate. ty!

justinfrench:
Closing due to lack of activity. Please feel free to re-open with new information!

wasifhossain:
we had to go slightly manual to integrate Reform into Activeadmin.

in order to **select the default value in the dropdown**, the following code would work WITHOUT Reform:

```rb
# app/admin/post.rb
# inside some action
post.author ||= Author.first

# app/views/admin/posts/_form.html.erb
<%= f.input :author, as: :select, collection: Author.pluck(:name, :id) %>
```

WITH Reform,
```rb
# app/forms/admin/posts/post_form.rb
class PostForm < ::Reform::Form
  property :author_id, prepopulator: ->(_options) { self.author_id ||= Author.first.id }
end

# app/admin/post.rb
# inside some action
post_form = PostForm.new(post)
post_form.prepopulate!

# app/views/admin/posts/_form.html.erb
<%= f.input :author_id, as: :select, collection: Author.pluck(:name, :id) %>
```

**Versions**:

Ruby 2.7.2
Rails 6.0.3.4
ActiveAdmin 2.8.1
Formtastic 4.0.0.rc1
Reform 2.3.3

PR #1142, opened 2015-03-19
Discover fields on a form object not just ActiveRecord

fernandes:
WIP: Need to check the [Reform PR](https://github.com/apotonick/reform/pull/212), then I add tests and doc before merge, just opened this PR so we can discuss seeing some code.

This PR is related to https://github.com/justinfrench/formtastic/issues/1141

My proposal to solve the problem on the issue, without affecting the behavior with AR models, so this way I just add a method on Reform to expose the fields key, then formtastic can consume on its InputsHelper

//cc @apotonick

### DIFF
--- a/lib/formtastic/helpers/inputs_helper.rb
+++ b/lib/formtastic/helpers/inputs_helper.rb
@@ -310,6 +310,7 @@ module Formtastic
       def default_columns_for_object
         cols  = association_columns(:belongs_to)
         cols += content_columns
+        cols += form_object_fields
         cols -= Formtastic::FormBuilder.skipped_columns
         cols.compact
       end
@@ -331,6 +332,14 @@ module Formtastic
         end
       end
       
+      # Collects fields from a form object that responds to this method
+      # If you'd like to implement on your form object just need to return a array of symbols
+      # [:field1, :field2]
+      def form_object_fields
+        return [] unless @object.respond_to?(:form_object_fields)
+        @object.form_object_fields
+      end
+
       # Collects association columns (relation columns) for the current form object class. Skips
       # polymorphic associations because we can't guess which class to use for an automatically
       # generated input.


apotonick:
Aaaah, now I understand! @fernandes added `#form_object_fields` to formtastic. I had a similar discussion with my friend @rafaelfranca where we started to define an API for form builders.

This looks exactly like this. In Reform, we started to collect/document methods required by simple_form here: https://github.com/apotonick/reform/blob/master/lib/reform/form/model_reflections.rb

I am guessing that formtastic requires the same methods (e.g. `#column_for_attribute`, etc.), right, @justinfrench?

Regarding this particular PR: I find it a good idea of @fernandes to expose a method that returns all fields from a "model". Right now, every form gem has its own way of collection this, so something like `#fields` would be beneficial for all form builders.

justinfrench:
@apotonick Formtastic was born out of ActiveRecord and adapted to ActiveModel and other ORMs over time. So yeah, `column_for_attribute` features heavily, as does a lot of column reflection code. I have no issue getting better at supporting form objects. We could likely just copy-paste whatever you work out with simple_form.

justinfrench:
@fernandes this looks okay. Would love the comments to discuss _why_ a bit (e.g. to support Reform), and obviously it needs test coverage and documentation additions as necessary before it could be merged.

fernandes:
thank you guys for your comments... :wink: 

@apotonick I have no expertise to create a common interface for all form builders / form objects to follow, but if @rafaelfranca would like to add something like this into rails I can give a shot

@justinfrench as you told formtastic needs to get the fields for AR/AM to make all this column reflection code, but Reform does not expose all its fields. As we (as ruby/rails community) don't have a common interface for form builders, I added a method to expose fields on Reform and here to discover... tested on ActiveAdmin+ActiveTrail and its working perfectly

if both you guys agree its a good solution I improve the PRs with tests and docs so we can get this integrations done... can I go ahead or you suggest anything else?

justinfrench:
@fernandes go for it!

justinfrench:
@fernandes any update?

fernandes:
hey @justinfrench , thank you for follow up, nobody decided about how forms and form objects could work together so I have made no advance. thank you

justinfrench:
Ok, closing for now, happy for it to be re-opened :)

fernandes:
:smile:

PR #1148, opened 2015-05-21
Checkboxes with collection syntax

### LINKED ISSUE #1125: Checkboxes with new collection syntax
Hello,

I am working on implementing the new collections version of the checkbox's api. In the past I was using the member label syntax to set the labels for each checkbox like so:

```
<%= f.input :permissions, :as => :check_boxes, member_label: :label_display %>
```

I switched to the new collections syntax. 
```
<%= f.input :permissions, :as => :check_boxes,
 collection: Permission.all.inject([]) { |coll, item| coll.push([item.label_display, item.id]) } %>
```
This syntax was not properly set the checked option on items that have been selected when editing an existing entry.

I also tried switching the to_s on the permissions to see if it would use the to string method for the name. I also tried setting all the label_method options I could find in the base class. This has lead me to needing to use the member_label in order to get the label I want and the properly selected checkboxes. But this has a deprecation warning. 

I am using formtastic in active admin. 

Thanks for the help,
Cory

gwincr11:
This commit resolves #1125 .
I placed a check to see if the array item is an active record
asset. If it is I get it's id, this properly creates an
array of id numbers to check against.

Warning, I could not get the tests to run on my computer.

### DIFF
--- a/lib/formtastic/inputs/check_boxes_input.rb
+++ b/lib/formtastic/inputs/check_boxes_input.rb
@@ -180,11 +180,19 @@ module Formtastic
       def make_selected_values
         if object.respond_to?(method)
           selected_items = object.send(method)
-
           # Construct an array from the return value, regardless of the return type
           selected_items = [*selected_items].compact.flatten
 
-          [*selected_items.map { |o| send_or_call_or_object(value_method, o) }].compact
+          #[*selected_items.map { |o| send_or_call_or_object(value_method, o) }].compact
+          selected = []
+          selected_items.each do |o|
+            if o.is_a? ActiveRecord::Base
+              selected << o.id
+            else
+              selected << send_or_call_or_object(value_method, 0)
+            end
+          end
+          selected
         else
           []
         end
--- a/lib/formtastic/inputs/check_boxes_input.rb
+++ b/lib/formtastic/inputs/check_boxes_input.rb
@@ -184,14 +184,10 @@ module Formtastic
           selected_items = [*selected_items].compact.flatten
 
           selected = []
-          selected_items.each do |o|
-            item = send_or_call_or_object(value_method, o)
-            if item == nil
-              item = o.id if o.respond_to? :id
-            end
-            selected << item
-          end
-          selected
+          selected_items.map do |selected_item|
+            selected_item_id = selected_item.id if selected_item.respond_to? :id
+            item = send_or_call_or_object(value_method, selected_item) || selected_item_id
+          end.compact
         else
           []
         end
--- a/lib/formtastic/inputs/check_boxes_input.rb
+++ b/lib/formtastic/inputs/check_boxes_input.rb
@@ -183,14 +183,13 @@ module Formtastic
           # Construct an array from the return value, regardless of the return type
           selected_items = [*selected_items].compact.flatten
 
-          #[*selected_items.map { |o| send_or_call_or_object(value_method, o) }].compact
           selected = []
           selected_items.each do |o|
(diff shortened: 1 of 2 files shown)


justinfrench:
@gwincr11 thanks! We'll need to work up a test case before this is merged. What issues were you having with the specs?

lib/formtastic/inputs/check_boxes_input.rb
```diff
@@ -180,11 +180,18 @@ def checkbox_input(choice)
...
-          [*selected_items.map { |o| send_or_call_or_object(value_method, o) }].compact
+          selected = []
+          selected_items.each do |o|
+            item = send_or_call_or_object(value_method, o)
+            if item == nil
+              item = o.id if o.respond_to? :id
+            end
+            selected << item
+          end
+          selected
```
tute: Syntax: this could be written as

```ruby
selected_items.map do |selected_item|
  selected_item_id = selected_item.id if selected_item.respond_to?(:id)
  send_or_call_or_object(value_method, selected_item) || selected_item_id
end
```

tute: And I assume we still need the `.compact` call at the end, to filter out `nil`s?

gwincr11: Yes, I will test these things out. I also need to add some test coverage for HABTM. I got the tests running on my machine by running bundle exec rspec but bundle exec rake still fails.

@justinfrench you seem to be heavily mocking and stubbing in your tests. Do you have an example of how you would want a HABTM test setup? I can make something from scratch but if you have a helper method or pattern you want followed it would make life a little easier.

justinfrench: @gwincr11 I know the tests are a little old school and a mess :( I think https://github.com/justinfrench/formtastic/blob/master/spec/inputs/check_boxes_input_spec.rb#L336-L359 would be the right place to add in more contexts or scenarios. At least that's where'd I'd start copying-and-pasting from.

gwincr11:
Just so you are aware, I added the requested tests.

spec/inputs/check_boxes_input_spec.rb
```diff
@@ -358,6 +358,31 @@
...
+      @output_buffer = ''
+      mock_everything
+
+      concat(semantic_form_for(@freds_post) do |builder|
+        concat(builder.input(:authors, as: :check_boxes, collection: Author.all))
+      end)
+    end
+
+    it 'should render checkboxes' do
+      # I'm aware these two lines test the same thing
```
justinfrench: Rather than noting that you're aware, could you update the comment to discuss _why_ you're doubt-testing? What's the rationale behind it? Could you instead check that `::Author.all.size` is a meaningful number and remove the appearance of double-testing?

gwincr11: This is actually just a copy and paste from your line above. I am not sure why who ever wrote the original tests doubted the testing. I just left it in because I don't know why it should be doubted, so I cannot safely remove it.

justinfrench: heh, okay

justinfrench:
Merging, thanks!

spec/inputs/check_boxes_input_spec.rb
gwincr11: I wish I had a better answer, but I am a bit ignorant as to this testing methodology. Sorry :(

gwincr11:
Awesome, thank you :)

PR #1158, opened 2015-07-10
Add auto readonly

st0012:
ActiveRecord has a method called [attr_readonly](http://apidock.com/rails/ActiveRecord/Base/attr_readonly/class), which can set attributes unupdatable. So I think it would be nice if formtastic can detect those readonly attributes and set their input readonly.(SimpleForm already did that)

### DIFF
--- a/lib/formtastic/inputs/base/validations.rb
+++ b/lib/formtastic/inputs/base/validations.rb
@@ -193,8 +193,7 @@ module Formtastic
         end
 
         def readonly?
-          opt_readonly = options[:input_html] && options[:input_html][:readonly]
-          opt_readonly || readonly_attribute?
+          readonly_from_options? || readonly_attribute?
         end
 
         def readonly_attribute?
@@ -202,6 +201,10 @@ module Formtastic
           object_class.respond_to?(:readonly_attributes) &&
             self.object.persisted? &&
             object_class.readonly_attributes.include?(column.name.to_s)
+        end
+
+        def readonly_from_options?
+          options[:input_html] && options[:input_html][:readonly]
         end
       end
     end
--- a/lib/formtastic/inputs/base/validations.rb
+++ b/lib/formtastic/inputs/base/validations.rb
@@ -192,6 +192,16 @@ module Formtastic
           validation_limit || column_limit
         end
 
+        def readonly?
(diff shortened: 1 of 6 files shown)


lib/formtastic/inputs/base/validations.rb
```diff
@@ -192,6 +192,17 @@ def limit
           validation_limit || column_limit
         end
 
+        def readonly?
+          opt_readonly = options[:input_html] && options[:input_html][:readonly]
```
justinfrench: It'd be nice to see this extracted to a method like `readonly_from_options?`

st0012: Sure, let me change it. Thanks.

spec/inputs/country_input_spec.rb
```diff
@@ -56,7 +56,7 @@
       priority_countries = ["Foo", "Bah"]
       semantic_form_for(@new_post) do |builder|
         builder.stub(:country_select).and_return(Formtastic::Util.html_safe("<select><option>...</option></select>"))
-        builder.should_receive(:country_select).with(:country, priority_countries, {}, {:id => "post_country", :required => false, :autofocus => false}).and_return(Formtastic::Util.html_safe("<select><option>...</option></select>"))
+        builder.should_receive(:country_select).with(:country, priority_countries, {}, {:id => "post_country", :required => false, :autofocus => false, :readonly => false}).and_return(Formtastic::Util.html_safe("<select><option>...</option></select>"))
```
justinfrench: Why did these existing tests need to change? Was it a quick hack to get the tests to pass, or is it because the `:readonly` option will always be passed from now on?

st0012: It's because the `:readonly` option will always be passed.

justinfrench:
This is cool! I've dropped a few comments in. Could you also document the new option and it's default behaviour in `InputHelper#input` where the shared option docs tend to go?

st0012:
Sure. Put it in  `@option options :input_html`(line 136) section, as a additional info?

justinfrench:
@st0012 yes, somewhere with the other `@option` docs would be great.

st0012:
I already added in the input_html section 😃

justinfrench:
Merged, thanks.

Issue #1146, opened 2015-04-06
We should fallback to StringInput for unknown column types


justinfrench:
When trying to find the most suitable input type for a given attribute, we often look to the type of database field. The current implementation doesn't have a fallback, which means we're constantly playing catch-up to add explicit mappings for column types that ActiveRecord supports.

If you have a database column type of `foo`, our fallback is to try `FooInput` if there's no explicit mapping, leading to an `unknown input` exception.

Here's the relevant code:

``` ruby
# default is a :string, a similar behaviour to Rails' scaffolding.
def default_input_type(method, options = {}) # @private
  if @object
    return :select  if reflection_for(method)

    return :file    if is_file?(method, options)
  end

  if column = column_for(method)
    # Special cases where the column type doesn't map to an input method.
    case column.type
    when :string
      return :password  if method.to_s =~ /password/
      return :country   if method.to_s =~ /country$/
      return :time_zone if method.to_s =~ /time_zone/
      return :email     if method.to_s =~ /email/
      return :url       if method.to_s =~ /^url$|^website$|_url$/
      return :phone     if method.to_s =~ /(phone|fax)/
      return :search    if method.to_s =~ /^search$/
      return :color     if method.to_s =~ /color/
    when :integer
      return :select    if reflection_for(method)
      return :select    if enum_for(method)
      return :number
    when :float, :decimal
      return :number
    when :datetime, :timestamp
      return :datetime_select
    when :time
      return :time_select
    when :date
      return :date_select
    end

    # Try look for hints in options hash. Quite common senario: Enum keys stored as string in the database.
    return :select    if column.type == :string && options.key?(:collection)
    # Try 3: Assume the input name will be the same as the column type (e.g. string_input).
    return column.type
  else
    return :select    if options.key?(:collection)
    return :password  if method.to_s =~ /password/
    return :string
  end
end
```

The problem is this line:

``` ruby
return column.type
```

I think this needs to be `return :string`, but before that happens, we'd need to ensure we have explicit mappings from any column types we know about (anything previously caught by `return column.type`), so that `StringInput` is the fallback, instead of `WhateverInput`.

Here's the replacement code with one line change and a TODO for where the extra mappings would go:

``` ruby
# default is a :string, a similar behaviour to Rails' scaffolding.
def default_input_type(method, options = {}) # @private
  if @object
    return :select  if reflection_for(method)

    return :file    if is_file?(method, options)
  end

  if column = column_for(method)
[...]

PR #1165, opened 2015-08-27
Added hstore input.

ryanfox1985:
Hi,
I added Hstore input.
Regards.

### DIFF
--- a/lib/formtastic/inputs.rb
+++ b/lib/formtastic/inputs.rb
@@ -28,6 +28,7 @@ module Formtastic
     autoload :SelectInput
     autoload :StringInput
     autoload :TextInput
+    autoload :HstoreInput
     autoload :TimeInput
     autoload :TimePickerInput
     autoload :TimeSelectInput
--- a/lib/formtastic/inputs/hstore_input.rb
+++ b/lib/formtastic/inputs/hstore_input.rb
@@ -1,0 +1,48 @@
+module Formtastic
+  module Inputs
+
+    # Outputs a simple `<label>` with a `<textarea>` wrapped in the standard
+    # `<li>` wrapper. This is the default input choice for database columns of the `:text` type,
+    # but can forced on any text-like input with `:as => :text`.
+    #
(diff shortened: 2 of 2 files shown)


mikz:
This is exact copy of TextInput no? Then you can do `f.input :hstore, as: :text` and you are good.

justinfrench:
@ryanfox1985 what problem are you trying to solve with this patch? What will Formtastic users be able to do better now that this has been added? I'll close if I haven't heard back in a while, but would love to hear from you.

mikz:
This would not be needed, if Formtastic would just default to string (#1146), right?

justinfrench:
@mikz This is proposing something close to (or exactly like) a `TextInput` (which makes sense to me). If we defaulted to `StringInput` as in #1146, you'd end up with a tiny text field instead if a text area.

I have no thoughts on how/when/where a hstore input is useful (@ryanfox1985, can you tell us more about your use case?), but if it is, this would would make sense to merge regardless of what happens in #1146.

@ryanfox1985 if this *is* a direct copy of the `TextInput` implementation perhaps consider subclassing instead so that they don't accidentally diverge over time?

mikz:
@justinfrench sure. I was more thinking why is this proposing it. Imo there is no way of map columns to inputs globally. If there would be a way to map `:hstore` to `:text` in form builder settings, this PR would not be needed. 

And to me looks like this is exact copy of text input. So rather than copying/subclassing it, it should be adding line to the column to input mapping in [`default_input_type`](https://github.com/justinfrench/formtastic/blob/11b22d0c770cd6d734fb637189d666def56f166e/lib/formtastic/helpers/input_helper.rb#L257) method.

Anyway, we should hear what @ryanfox1985 tried to accomplish with this, so we can decide if it is the right way to do it or there are better things to do.

ryanfox1985:
Hi,

Sorry but when I created the PR I didn't know that you can render the
hstore fields as text input.

Regards.

PR #1169, opened 2015-10-11
Remove duplicate test

pranas:
I was adding a new spec when I noticed there are two identical specs. The exact same spec to the one I removed is above, [line 303](https://github.com/pranas/formtastic/blob/e0911a140c82eca586d9c0cff8466f9468b6ad8a/spec/inputs/select_input_spec.rb#L303).


mikz:
Thank you!

PR #1170, opened 2015-10-12
Add test-unit development dependency

pranas:
I was setting up development environment and got a crash when running on Ruby 2.2.3:

```
/path/to/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/rspec-rails-2.99.0/lib/rspec/rails/adapters.rb:17:in `require': cannot load such file -- test/unit/assertions (LoadError)
	from /path/to/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/rspec-rails-2.99.0/lib/rspec/rails/adapters.rb:17:in `rescue in <module:Rails>'
	from /path/to/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/rspec-rails-2.99.0/lib/rspec/rails/adapters.rb:12:in `<module:Rails>'
	from /path/to/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/rspec-rails-2.99.0/lib/rspec/rails/adapters.rb:6:in `<module:RSpec>'
	from /path/to/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/rspec-rails-2.99.0/lib/rspec/rails/adapters.rb:5:in `<top (required)>'
	from /path/to/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/ammeter-1.1.1/lib/ammeter/rspec/generator/example/generator_example_group.rb:3:in `require'
	from /path/to/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/ammeter-1.1.1/lib/ammeter/rspec/generator/example/generator_example_group.rb:3:in `<top (required)>'
	from /path/to/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/ammeter-1.1.1/lib/ammeter/rspec/generator/example.rb:2:in `require'
	from /path/to/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/ammeter-1.1.1/lib/ammeter/rspec/generator/example.rb:2:in `<top (required)>'
	from /path/to/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/ammeter-1.1.1/lib/ammeter/init.rb:7:in `require'
	from /path/to/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/ammeter-1.1.1/lib/ammeter/init.rb:7:in `<top (required)>'
	from /path/to/Development/formtastic/spec/spec_helper.rb:20:in `require'
	from /path/to/Development/formtastic/spec/spec_helper.rb:20:in `<top (required)>'
	from /path/to/Development/formtastic/spec/action_class_finder_spec.rb:2:in `require'
	from /path/to/Development/formtastic/spec/action_class_finder_spec.rb:2:in `<top (required)>'
	from /path/to/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/rspec-core-2.99.2/lib/rspec/core/configuration.rb:1065:in `load'
	from /path/to/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/rspec-core-2.99.2/lib/rspec/core/configuration.rb:1065:in `block in load_spec_files'
	from /path/to/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/rspec-core-2.99.2/lib/rspec/core/configuration.rb:1065:in `each'
	from /path/to/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/rspec-core-2.99.2/lib/rspec/core/configuration.rb:1065:in `load_spec_files'
	from /path/to/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/rspec-core-2.99.2/lib/rspec/core/command_line.rb:18:in `run'
	from /path/to/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/rspec-core-2.99.2/lib/rspec/core/runner.rb:103:in `run'
	from /path/to/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/rspec-core-2.99.2/lib/rspec/core/runner.rb:17:in `block in autorun'
```

Starting from Ruby 2.2.0 ``Test::Unit`` was removed from standard library. You got to add it to your Gemfile.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ rvm:
   - 1.9.3
   - 2.0.0
   - 2.1
+  - 2.2
 gemfile:
   - gemfiles/rails_3.2.gemfile
   - gemfiles/rails_4.gemfile
@@ -14,6 +15,8 @@ gemfile:
   - gemfiles/rails_edge.gemfile
 env:
   - DEFER_GC=false RAILS_EDGE=true
+before_install:
+  - gem install bundler -v 1.10.6
 script: "bundle exec rake spec"
 matrix:
   allow_failures:
@@ -24,5 +27,8 @@ matrix:
       gemfile: gemfiles/rails_edge.gemfile
       env: DEFER_GC=false RAILS_EDGE=true
     - rvm: 2.1
+      gemfile: gemfiles/rails_edge.gemfile
+      env: DEFER_GC=false RAILS_EDGE=true
+    - rvm: 2.2
       gemfile: gemfiles/rails_edge.gemfile
       env: DEFER_GC=false RAILS_EDGE=true
--- a/formtastic.gemspec
+++ b/formtastic.gemspec
@@ -28,14 +28,14 @@ Gem::Specification.new do |s|
   s.add_dependency(%q<actionpack>, [">= 3.2.13"])
 
   s.add_development_dependency(%q<nokogiri>)
-  s.add_development_dependency(%q<rspec-rails>, ["~> 2.14"])
+  s.add_development_dependency(%q<rspec-rails>, ["~> 3.3.2"])
   s.add_development_dependency(%q<rspec_tag_matchers>, ["~> 1.0"])
   s.add_development_dependency(%q<hpricot>, ["~> 0.8.3"])
   s.add_development_dependency(%q<RedCloth>, ["~> 4.2"]) # for YARD Textile formatting
   s.add_development_dependency(%q<yard>, ["~> 0.8"])
   s.add_development_dependency(%q<colored>, ["~> 1.2"])
   s.add_development_dependency(%q<tzinfo>)
-  s.add_development_dependency(%q<ammeter>, ["1.1.1"])
+  s.add_development_dependency(%q<ammeter>, ["~> 1.1.3"])
   s.add_development_dependency(%q<appraisal>, ["~> 1.0"])
   s.add_development_dependency(%q<rake>)
   s.add_development_dependency(%q<activemodel>, [">= 3.2.13"])
--- a/gemfiles/rails_3.2.gemfile
+++ b/gemfiles/rails_3.2.gemfile
@@ -4,4 +4,6 @@ source "https://rubygems.org"
 
 gem "rails", "~> 3.2.0"
 
+gem "test-unit-minitest", :platform => :ruby_22
+
 gemspec :path => "../"
--- a/lib/formtastic/inputs/country_input.rb
+++ b/lib/formtastic/inputs/country_input.rb
@@ -67,8 +67,10 @@ module Formtastic
     class CountryInput 
       include Base
 
+      CountrySelectPluginMissing = Class.new(StandardError)
+
       def to_html
-        raise "To use the :country input, please install a country_select plugin, like this one: https://github.com/stefanpenner/country_select" unless builder.respond_to?(:country_select)
+        raise CountrySelectPluginMissing, "To use the :country input, please install a country_select plugin, like this one: https://github.com/stefanpenner/country_select" unless builder.respond_to?(:country_select)
         input_wrapping do
           label_html <<
           builder.country_select(method, priority_countries, input_options, input_html_options)
--- a/spec/inputs/country_input_spec.rb
+++ b/spec/inputs/country_input_spec.rb
@@ -17,7 +17,7 @@ describe 'country input' do
         semantic_form_for(@new_post) do |builder|
           concat(builder.input(:country, :as => :country))
         end
-      }.should raise_error
+      }.should raise_error(Formtastic::Inputs::CountryInput::CountrySelectPluginMissing)
     end
 
   end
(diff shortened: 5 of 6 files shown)


mikz:
The problem with adding `test-unit` as a gem is that it is not what is in Ruby 2 stdlib.
Ruby has Test Unit stub built on minitest, where the test-unit gem is implementation from scratch.

In my projects I solved it in gemfile:

```ruby
platform :mri_22 do
  gem 'test-unit-minitest', require: false, group: :test
end
```

pranas:
If we add the gem it would override whatever is in Ruby 2 stdlib, no?

mikz:
If added to a gemspec, yes. Gemfile can target platform (MRI >= 2.2) so it uses it only on Ruby 2.2.
The `test-unit-minitest` gem is the test-unit implementation based on minitest that was in stdlib.

pranas:
Yeah, but is there any problem with using ``test-unit`` gem in earlier Ruby versions? I think it would be more consistent. I would avoid having platform specific cases if not necessary.

mikz:
I think it would be going backwards. Formtastic uses  rspec so it should not need test unit / minitest. Rails uses minitest, so it does not needs test unit. 
This would be adding 3rd testing framework, just to fulfill some ancient dependency. I'd much rather update rspec, to drop the test-unit dependency.

pranas:
I agree upgrading RSpec seems like the best solution. Are you aware of any difficulties to upgrade to RSpec 3? I could give it a try.

mikz:
According to the upgrade guide: https://relishapp.com/rspec/docs/upgrade
Just fixing the deprecations should be enough. And I don't see any in https://travis-ci.org/justinfrench/formtastic/jobs/85088684
So it might be seamless to 3.0.

pranas:
I upgraded to the latest RSpec but then RSpec itself suggested adding ``test-unit`` gem:

> Ruby 2.2+ has removed test/unit from the core library. Rails requires this as a dependency. Please add test-unit gem to your Gemfile: `gem 'test-unit', '~> 3.0'`

More details here: https://github.com/rspec/rspec-rails/blob/master/lib/rspec/rails/adapters.rb#L22

It's actually only needed in Ruby 2.2+ and Rails >= 3.2.21 combination. I guess we should add it while we support Rails 3.2. Your thoughts?

mikz:
@pranas Rails 3.2 can work fine with the minitest version of test/unit. I'd use the `test-unit-minitest` gem on Ruby 2.2 and only for Rails 3.2.
That way all the versions will run on minitest (as they were already from ruby 1.9).

pranas:
Sorry it took me so long to come back to this. I changed to ``test-unit-minitest`` for Rails 3.2 on Ruby 2.2 case and I upgraded to RSpec 3 to solve the Rails 4.0/4.1 on Ruby 2.2 situation. RSpec 3 throws a lot of deprecations. Not sure if or when I would be able to fix all of them. Can we get this merged in and deal with the deprecations later?

.travis.yml
```diff
@@ -14,6 +15,8 @@ gemfile:
   - gemfiles/rails_edge.gemfile
 env:
   - DEFER_GC=false RAILS_EDGE=true
+before_install:
+  - gem install bundler
```
mikz: would it make sense to install some specific version like 1.10.6 ?

spec/support/test_environment.rb
```diff
@@ -6,6 +6,7 @@
   config.include RspecTagMatchers
   config.include CustomMacros
   config.mock_with :rspec
+  config.deprecation_stream = '/dev/null'
```
mikz: I'm not sure. We will drop rails 3 but this will not show deprecations for rails 4 etc.

mikz:
@pranas for me it looks good! Had two minor comments, but after that it should be good to go. Thanks a lot!

spec/support/test_environment.rb
pranas: It's only RSpec deprecation stream, not Rails. While I don't like this either, the test suite generates a ton of deprecations because of old should syntax and other things.

.travis.yml
pranas: Yeah, it would. I'll lock it at 1.10.6.

spec/support/test_environment.rb
mikz: Ok. This still can be a surprise when someone wants to upgrade rspec. I'd not expect it to be silenced.

pranas: Ok, let's get rid of this and maybe someone sometime will fix them. I expect a lot can be done with regular expressions and find & replace.

pranas:
@mikz I made changes according to your comments.

mikz:
@pranas sorry for being pain. Thank you for awesome work!

PR #1171, opened 2015-10-12
Hint at "appraisal install" in contributing guide

pranas:
It wasn't obvious to me how to set up development environment for formtastic. I looked at contributing guide and it didn't help. I think a hint at ``appraisal install`` there would help.


mikz:
Thanks!

PR #1172, opened 2015-10-24
Improve timezone input

st0012:
Hi~ I added default time_zones configuration, updated documentation and some tests.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,3 +1,5 @@
 source 'https://rubygems.org'
 
 gemspec
+
+gem "pry"
--- a/Gemfile
+++ b/Gemfile
@@ -1,5 +1,3 @@
 source 'https://rubygems.org'
 
 gemspec
-
-gem "pry"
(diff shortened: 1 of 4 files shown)


st0012:
@mikz Do you have any thought about this PR?

mikz:
@st0012 I'm not much a fan of adding a global configuration for input specific thing.

Would it make sense for you to create a custom input instead?

st0012:
I added the global configuration because I see the `Add config` todo. And there already is the `priority_zones` option to set priority zone per input, so I think the `Add config` means global configuration.

mikz:
@st0012 there are already `priority_countries` in the configuration, so to be consistent it makes sense to add this!

st0012:
@mikz ok, thanks for reviewing!

justinfrench:
Looks good, looks consistent, looks backwards compatible! Merging, thanks!

st0012:
:smile:

PR #1174, opened 2015-10-25
Speed up some implementation

st0012:
Hello, I use [fast-ruby](https://github.com/JuanitoFatas/fast-ruby) as a reference, found some improvable code that can be speeded up. Related benchmarks can be found in commit messages.

### DIFF
--- a/lib/formtastic/helpers/fieldset_wrapper.rb
+++ b/lib/formtastic/helpers/fieldset_wrapper.rb
@@ -22,7 +22,7 @@ module Formtastic
       #   f.inputs :my_little_legend, :title, :body, :author    # Localized (118n) legend with I18n key => I18n.t(:my_little_legend, ...)
       #   f.inputs :title, :body, :author                       # First argument is a column => (no legend)
       def field_set_and_list_wrapping(*args, &block) # @private
-        contents = args.last.is_a?(::Hash) ? '' : args.pop.flatten
+        contents = args[-1].is_a?(::Hash) ? '' : args.pop.flatten
         html_options = args.extract_options!
 
         if block_given?
@@ -57,20 +57,20 @@ module Formtastic
       def parent_child_index(parent) # @private
         # Could be {"post[authors_attributes]"=>0} or { :authors => 0 }
         duck = parent[:builder].instance_variable_get('@nested_child_index')
-        
+
         # Could be symbol for the association, or a model (or an array of either, I think? TODO)
         child = parent[:for]
         # Pull a sybol or model out of Array (TODO: check if there's an Array)
         child = child.first if child.respond_to?(:first)
         # If it's an object, get a symbol from the class name
         child = child.class.name.underscore.to_sym unless child.is_a?(Symbol)
-        
+
         key = "#{parent[:builder].object_name}[#{child}_attributes]"
 
-        # TODO: One of the tests produces a scenario where duck is "0" and the test looks for a "1" 
+        # TODO: One of the tests produces a scenario where duck is "0" and the test looks for a "1"
         # in the legend, so if we have a number, return it with a +1 until we can verify this scenario.
         return duck + 1 if duck.is_a?(Fixnum)
-        
+
         # First try to extract key from duck Hash, then try child
         (duck[key] || duck[child]).to_i + 1
       end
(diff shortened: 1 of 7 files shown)


mikz:
:+1:

justinfrench:
@st0012

Interesting, I learned something :)

Subjectively, I think the code proposed here is not as clear as the code it replaces. I personally have to _think_ more about what `[-1]` and `yield` are doing compared with `last` and `call`. Others may not agree, of course.

So, we're making a trade-off of "less readable code" for "faster code", but it's not clear to me if this code would be _meaningfully_ faster (in an average full Rails render with a Formtastic form).

Did you happen to run any benchmarks on your app?

st0012:
@justinfrench 

Ok, I will try to figure out a way to benchmark these changes. However, I think `yield` is a pretty normal way to execute a block, so in my opinion the only `trade-off` is the `last` changed to `[-1]`.

st0012:
@mikz @justinfrench 

Hello, it's me again. I did a little benchmark to compare array methods `last` and `[-1]`'s effect in formtastic.

Here's the benchmark code, you can also find it in my fork's [PR](https://github.com/st0012/formtastic/pull/2)
```ruby
module Formtastic
  module Inputs
    module Base
      module FastCollections
        include Collections
        def value_method
          @value_method ||= (value_method_from_options || label_and_value_method[-1])
        end

        def label_and_value_method_from_collection(_collection)
          sample = _collection.first || _collection[-1]

          case sample
          when Array
            label, value = :first, :last
          when Integer
            label, value = :to_s, :to_i
          when Symbol, String, NilClass
            label, value = :to_s, :to_s
          end

          # Order of preference: user supplied method, class defaults, auto-detect
          label ||= builder.collection_label_methods.find { |m| sample.respond_to?(m) }
          value ||= builder.collection_value_methods.find { |m| sample.respond_to?(m) }

          [label, value]
        end
      end
    end
  end
end

module Formtastic
  module Inputs
    class FastSelectInput < SelectInput
      include Base::FastCollections
    end
  end
end

class ViewCreator
  include Formtastic::Helpers::FormHelper
  include FormtasticSpecHelper
  include RSpec::Mocks
  Formtastic::FormBuilder.input_class_finder = Formtastic::InputClassFinder

  def double(name=nil, stubs={})
    RSpec::Mocks::Double.new(name, stubs)
  end

  def create_selection_1
    mock_everything
    semantic_form_for(@new_post, url: "/hello") do |builder|
      builder.input(:title, :as => :select, :collection => ["Title A", "Title B", "Title C"])
    end
  end

  def create_selection_2
    mock_everything
    semantic_form_for(@new_post, url: "/hello") do |builder|
      builder.input(:title, :as => :fast_select, :collection => ["Title A", "Title B", "Title C"])
    end
  end
end

view = ViewCreator.new

Benchmark.ips do |x|
  x.report("create_selection") { view.create_selection_1 }
  x.report("create_fast_selection") { view.create_selection_2 }
  x.compare!
end
```

And the result is

```
Calculating -------------------------------------
    create_selection     2.000  i/100ms
create_fast_selection
                         3.000  i/100ms
-------------------------------------------------
    create_selection     31.698  (±18.9%) i/s -    152.000
create_fast_selection
                         32.072  (±18.7%) i/s -    153.000

Comparison:
create_fast_selection:       32.1 i/s
    create_selection:       31.7 i/s - 1.01x slower
```

However, I will give it more tries later, hope that I can reduce the errors.

st0012:
Here's the benchmark of `[-1]` and `last`'s difference in generating form. But I'm not sure if `1.07x` slower is a big difference. What do you think? Would you mind run the benchmark again to verify it?
cc @mikz @justinfrench

mikz:
I don't know. I'm totally ok with the change, even though the impact on the real usage is minimal.
There might be another hotspots what would yield bigger benefits. 
I'm fine with it because for me `[-1]` might be more clearer than `.last` but that is just my personal feeling. 
The `[]` form is not so spread across ruby land so it might be confusing to some people. Also, it might be a compiler problem so this just should be equally fast and might be fixed in future versions of ruby.

justinfrench:
I basically agree with @mikz, but I don't see any real harm in merging. 

@st0012 final check — have you used your fork in a real app and/or production code? I'm a little bit nervous on the `block.call()` vs `yield` change. Real world use + free build would make me happier than green build alone ;)

st0012:
I have tested it in my current working project and uses it with activeadmin. I can create/edit record's nested attributes normally without any issue. And `yield` is a very common usage that often compared with `block.call` like [this](http://stackoverflow.com/questions/1410160/ruby-proccall-vs-yield). So I think the change should makes no functional difference :)

mikz:
I totally agree with the `yield`. Allocating a Proc for every method call can be expensive. Not only for time but also for memory.

justinfrench:
Merging, thanks!

PR #1178, opened 2015-11-06
refactoring with rubocop to current rails standards

Bobby-H:


mikz:
I agree we should unify the codebase and apply some rubocop config. However this PR addresses just one file.

mikz:
There is different PR that tries to push some rubocop rules #1209.

justinfrench:
Closing as there's a more complete PR in #1209

PR #1179, opened 2015-11-13
Support validation contexts as Array

McRip:
Hello Justin,

i added support for displaying a field as required, when the validation context in the model is specified as an array to have multiple validation contexts for a validation.
I would be very happy if this could be merged.

Thanks in advance
Moritz

### DIFF
--- a/lib/formtastic/inputs/base/validations.rb
+++ b/lib/formtastic/inputs/base/validations.rb
@@ -138,7 +138,8 @@ module Formtastic
           if validations?
             validations.select { |validator|
               if validator.options.key?(:on)
-                return false if (validator.options[:on] != :save) && ((object.new_record? && validator.options[:on] != :create) || (!object.new_record? && validator.options[:on] != :update))
+                validator_on = Array(validator.options[:on])
+                return false if (validator_on.exclude?(:save)) && ((object.new_record? && validator_on.exclude?(:create)) || (!object.new_record? && validator_on.exclude?(:update)))
               end
               case validator.kind
               when :presence
--- a/spec/inputs/base/validations_spec.rb
+++ b/spec/inputs/base/validations_spec.rb
@@ -1,0 +1,294 @@
+require 'fast_spec_helper'
+require 'inputs/base/validations'
+
+class MyInput
+  include Formtastic::Inputs::Base::Validations
+end
+
+describe MyInput do
+  let(:builder) { double }
+  let(:template) { double }
+  let(:model_class) { double }
+  let(:model) { double(:class => model_class) }
+  let(:model_name) { "post" }
+  let(:method) { double }
+  let(:options) { Hash.new }
+  let(:validator) { double }
+  let(:instance) { MyInput.new(builder, template, model, model_name, method, options) }
+
+  describe '#required?' do
+    before :each do
+      allow(instance).to receive(:validations?).and_return(:true)
+      allow(instance).to receive(:validations).and_return([validator])
+    end
+
+    context 'with options[:required] being true' do
+      let(:options) { {required: true} }
+
+      it 'is required' do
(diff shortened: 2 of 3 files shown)


lib/formtastic/inputs/base/validations.rb
```diff
@@ -138,7 +138,11 @@ def required?
           if validations?
             validations.select { |validator|
               if validator.options.key?(:on)
-                return false if (validator.options[:on] != :save) && ((object.new_record? && validator.options[:on] != :create) || (!object.new_record? && validator.options[:on] != :update))
+                if validator.options[:on].is_a?(Array)
+                  return false if (validator.options[:on].exclude?(:save)) && ((object.new_record? && validator.options[:on].exclude?(:create)) || (!object.new_record? && validator.options[:on].exclude?(:update)))
```
mikz: Could this be simplified by always converting it to an Array like `Array(validator.options[:on])` and then just checking for `exclude?` ?
Having similar conditions in two branches is really dangerous.

st0012: I think this can be more simple because you don't actually need that `if` statement, just return the opposite value will do the job
```ruby
!(validator.options[:on].exclude?(:save) &&
  ((object.new_record? && validator.options[:on].exclude?(:create)) ||
   (!object.new_record? && validator.options[:on].exclude?(:update))
))
```
And you can also create an private method so that the condition can be more readable
```ruby
def exclude_action?(action)
  validator.options[:on].exclude?(action)
end
```

```ruby
!(exclude_action?(:save) &&
  ((object.new_record? && exclude_action?(:create)) ||
   (!object.new_record? && exclude_action?(:update))
))
```

Just some advise :smile:

McRip:
Thank you for your comments. You were right, i was a bit quick on this. I prefer @mikz solution to just convert it to an array and then have only one check left. I changed my commit.

lib/formtastic/inputs/base/validations.rb
```diff
@@ -138,7 +138,8 @@ def required?
           if validations?
             validations.select { |validator|
               if validator.options.key?(:on)
-                return false if (validator.options[:on] != :save) && ((object.new_record? && validator.options[:on] != :create) || (!object.new_record? && validator.options[:on] != :update))
+                validator_on = validator.options[:on].is_a?(Array) ? validator.options[:on] : [validator.options[:on]]
```
mikz:
```ruby
validator.options[:on].is_a?(Array) ? validator.options[:on] : [validator.options[:on]]
```
is the same as
```ruby
Array(validator.options[:on])
```
`Array` is also a function. See http://docs.ruby-lang.org/en/2.2.0/Kernel.html#method-i-Array.

McRip:
Thank you @mikz, you are so right. I changed my implementation and pushed a new version of my PR.
@justinfrench: Can this be merged?

mikz:
@McRip would it be possible to have a real unit test for that logic? Because the tests you added cover the case when you add an array, but does not test what happens when you have there both `:save` and `:create` or different combinations. Testing this on the 'integration' level is good for some examples, but the edge cases could be tested on lower level which would reduce amount of tests and noise they add. But that is just my 2c. Just a nitpick.

McRip:
@mikz I created unit tests for Formtastic::Inputs::Base::Validations#required?. I hope this can be merged soon, so that i can ditch my git dependency in my Gemfile :)

justinfrench:
@mikz Looks good to me. I'll merge in a couple of days, but hold off in case you have additional comments.

mikz:
@McRip wow! Very nice :+1: 
@justinfrench looks great to me!

justinfrench:
Thanks @McRip!

McRip:
Thanks for merging!

Issue #1183, opened 2016-01-05
Rails 5 beta1 support


marcomd:
Hello,

i'm testing the current master with rails5 beta1, ruby 2.2.3 mingw x64 as i just reported in the [activeadmin issue #4177](https://github.com/activeadmin/activeadmin/issues/4177)

```
Formtastic::UnknownInputError - Unable to find input class Input:
   () ruby/ruby-223x64p173/lib/ruby/gems/2.2.0/bundler/gems/formtastic-88f28a0bc2cd/lib/formtastic/helpers/input_helper.rb:335:in `rescue in namespaced_input_class'
```

```ruby
      def namespaced_input_class(as)
        @input_class_finder ||= input_class_finder.new(self)
        @input_class_finder.find(as)
      rescue Formtastic::InputClassFinder::NotFoundError
        raise Formtastic::UnknownInputError, "Unable to find input #{$!.message}"
      end
```

mikz:
@marcomd could you provide a sample app and a steps/test to reproduce it?

I'll try to enable Rails 5 build.

marcomd:
rails new Foo -m https://www.dropbox.com/s/7m6wcshcfv0qrc6/active_template_rails5.rb

Do you use git ? **n**
Do you want to make development easier? **n**
Add rspec as testing framework ? **n**
Authentication ? **y**
 Insert model name: [User] => **enter**
Authorization ? **y**
Do you have to handle states ? **n**
Would you use dashboard as root ? (recommended) **y**
Insert folder name to install locally: [blank=default gems path] => **enter**

cd Foo
rails s
Go to http://localhost:3000/admin/login
I get the error above

seanlinsley:
This is a problem with Devise, because it isn't using the new attributes API in Rails 5: https://github.com/plataformatec/devise/issues/3880

seanlinsley:
Here's a full stack trace: https://github.com/activeadmin/activeadmin/pull/4254#issuecomment-168422718

jtomaszewski:
Do you guys have an idea for any workaround to make it work on Rails5, before devise starts using the new attributes API ?

jtomaszewski:
#1184 makes it work correctly on Rails 5. If someone wants to make it work on rails 5, you can use my patch-1 branch or wait for this commit to be merged.

mikz:
Fixed by #1184

jtomaszewski:
Thanks for quick reaction ;)

mikz:
[...]

PR #1184, opened 2016-01-11
Treat input as :string if attached `column.type` is empty

jtomaszewski:
`#column_for` can return `ActiveRecord::ConnectionAdapters::NullColumn`. `column.type` will be `nil` then.
This fix makes the InputHelper#default_input_type recognize such column as string by default, just as the whole `column` would be `nil`.

Fixes the `Unable to find input class Input` error (justinfrench/formtastic#1183).


mikz:
Looks good!

PR #1188, opened 2016-01-22
[rails5] support for rails 5 attributes api

mikz:
This adds experimental support for Rails 5 Attributes API.

## TODO

- [x] integration tests (new project, install formtastic, add a model, generate a form, ...) (done in #1205)
- [x] integration test for rails 5

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,5 +31,4 @@ matrix:
     - rvm: 2.1.8
       gemfile: gemfiles/rails_edge.gemfile
   allow_failures:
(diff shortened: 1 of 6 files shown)


mikz:
@justinfrench wdyt? I'd say it is done. It adds support for [Rails 5 Attributes API](http://edgeapi.rubyonrails.org/classes/ActiveRecord/Attributes/ClassMethods.html#method-i-attribute).
The integration test overrides the type, and checks if it renders correctly.  On Rails 4 it should work as it used to.

justinfrench:
:+1:

PR #1190, opened 2016-01-31
Try to fix the contradict between the hidden_fields for each checkbox…

### LINKED ISSUE #1189: Shouldn't it be showing a hidden input if `hidden_fields` is set to be true?
The line https://github.com/justinfrench/formtastic/blob/master/lib/formtastic/inputs/check_boxes_input.rb#L102 indicates that an empty string will be returned if `hidden_fields` is set to be true, while a hidden input will be returned if it's set to be false.

While the line https://github.com/justinfrench/formtastic/blob/master/lib/formtastic/inputs/check_boxes_input.rb#L173 tells a different story, it will return a hidden input if the `hidden_fields` is set to be true.

It contradicts between each other if I understand it correctly.

leomayleomay:
… and the hidden_fields for all

Closes #1189

### DIFF
--- a/lib/formtastic/inputs/check_boxes_input.rb
+++ b/lib/formtastic/inputs/check_boxes_input.rb
@@ -100,12 +100,12 @@ module Formtastic
 
       def hidden_field_for_all
         if hidden_fields?
-          ''
-        else
           options = {}
           options[:class] = [method.to_s.singularize, 'default'].join('_') if value_as_class?
           options[:id] = [object_name, method, 'none'].join('_')
           template.hidden_field_tag(input_name, '', options)
+        else
+          ''
         end
       end
 
--- a/spec/inputs/check_boxes_input_spec.rb
+++ b/spec/inputs/check_boxes_input_spec.rb
@@ -56,8 +56,8 @@ describe 'check_boxes input' do
       output_buffer.should_not have_tag("form li fieldset ol li input[@type='hidden']")
     end
 
-    it 'should render one hidden input for each choice outside the ol' do
-      output_buffer.should have_tag("form li fieldset > input[@type='hidden']", :count => 1)
+    it 'should not render hidden input for each choice outside the ol' do
+      output_buffer.should_not have_tag("form li fieldset > input[@type='hidden']")
     end
 
     describe "each choice" do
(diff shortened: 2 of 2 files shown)


mikz:
I'd say it looks good. Could you add a test? It is apparently not covered, because the build is green.

leomayleomay:
@mikz actually, I didn't add any new specs for this, I just changed several existing ones, :), https://github.com/leomayleomay/formtastic/commit/10d9ebc3ad56c9d7f3a5290ec4b6be5114efa61c

mikz:
@leomayleomay sorry! Totally missed it!
Looks good I guess. Took a while to understand it and it looks correct.

leomayleomay:
@mikz I am wondering when this will gets merged? So I can close the issue #1189, cheers

mikz:
This will definitely require a changelog entry. How you want to proceed @justinfrench ?

justinfrench:
Wow — that's a strange bug/contradiction!

Even though this is arguably a bug fix, it's been that way for so long and feels substantial enough that it should be considered a change in behaviour. So I'd prefer to see it merged into 4.0-dev.

If you disagree, I'm happy for it to go into master for a 3.x, though it'd probably need to be a 3.2 instead of 3.1.x now.

What do you think @mikz?

mikz:
I'd be in for merging it into 4.0.  I don't know how we could make it opt in in 3.2 and default in 4.0 as it is just a boolean param. But it looks like a very fringe feature I had no idea that exists, so imho change in 4.0 would be fine.

mikz:
CHANGELOG: Fixed checkboxes hidden_fields contradiction. Now shows hidden fields when hidden_fields option is true.

mikz:
@leomayleomay thank you for the effort! This will be released in Formtastic 4. You can track the progress in https://github.com/justinfrench/formtastic/issues/1197.

justinfrench:
Thanks @leomayleomay & @mikz!

leomayleomay:
my pleasure  
  
Sent from [Nylas N1](https://nylas.com/n1?ref=n1), the extensible, open source
mail client.  

On Mar 27 2016, at 3:47 pm, Justin French &lt;notifications@github.com&gt;
wrote:  

> Thanks [@leomayleomay](https://github.com/leomayleomay) &amp;
[@mikz](https://github.com/mikz)!

>

> —  
You are receiving this because you were mentioned.  
Reply to this email directly or [view it on GitHub](https://github.com/justinf
rench/formtastic/pull/1190#issuecomment-201977176)

PR #1191, opened 2016-02-29
Bugfix for multiple validations with validation contexts

McRip:
Hello,

while working with validation contexts i found a bug where a form field was not marked as required even if a presence validator without any options was attaches to this field.
This occured because the validation contexts check returned false if some conditions were not met. This caused other validations not being evaluated.
This PR fixes this behaviour and adds test for it. I hope this can be merged.

Cheers,
Moritz


mikz:
Looks good to me!

mikz:
I'm merging this as 3.1.4 release will be soon and this should be in!

PR #1193, opened 2016-03-23
Fixed inclusion validations not being correctly marked as optional

markokajzer:
#### Current behavior
Setting allow_blank to true on an inclusion validation correctly marks it as non-required.
Setting allow_nil to true however, does not have any influence on whether the attribute is marked as required or not.

#### New behavior
Setting allow_nil to true will correctly mark the attribute as non-required.


mikz:
If the field is in the form. It will be send as a parameter. And it will be in rails as blank, not as nil. So, if you have allow_nil: true, but not allow_blank: true, it would still be required. Correct?

markokajzer:
Yeah, you are right. I was working with enums and totally overlooked everything else! :smiley:

PR #1195, opened 2016-03-26
[travis] improve build

mikz:
* remove env variables that were not used
* cleanup travis.yml
* allow Ruby 1.9.3 failures (as some gems fail to install now)


mikz:
Merging this. Master is red because 1.9.3 fails to install some dependencies in some cases. 

1.9.3 EOL was more than a year ago https://www.ruby-lang.org/en/news/2014/01/10/ruby-1-9-3-will-end-on-2015/

That does not mean we don't support it. Some Rails versions pass.

Issue #911, opened 2013-01-15
Formtastic::UnknownInputError: Unable to find input class StringInput


vfilesgroup:
Hi there,

From time to time on our heroku deployed app we see this error (sometimes with TextInput, too). Restarting the dynos fixes it for a while, but it would be nice to get to the bottom of this. Has anyone else seen this, or have a fix for it?

Thanks!

xlymian:
We are encountering the same issue sometimes with BooleanInput, but more often with HiddenInput. Here is small log extract:

```
2015-05-18T11:45:12.629297+00:00 app[web.3]: ActionView::Template::Error (Unable to find input class HiddenInput):
2015-05-18T11:45:12.629299+00:00 app[web.3]:     13:     = semantic_form_for Problem.new(current_user, params, request.original_url) do |f|
2015-05-18T11:45:12.629301+00:00 app[web.3]:     14:       = f.semantic_errors
2015-05-18T11:45:12.629303+00:00 app[web.3]:     15:       = f.inputs do
2015-05-18T11:45:12.629305+00:00 app[web.3]:     16:         = f.input :controller_name, as: :hidden
2015-05-18T11:45:12.629306+00:00 app[web.3]:     17:         = f.input :action_name, as: :hidden
2015-05-18T11:45:12.629308+00:00 app[web.3]:     18:         = f.input :current_user_id, as: :hidden
2015-05-18T11:45:12.629310+00:00 app[web.3]:     19:         = f.input :params, as: :hidden
2015-05-18T11:45:12.629312+00:00 app[web.3]:   app/views/problems/_new.html.haml:16:in `block (2 levels) in _app_views_problems__new_html_haml__3493347280847073492_69899255027720'
2015-05-18T11:45:12.629314+00:00 app[web.3]:   app/views/problems/_new.html.haml:15:in `block in _app_views_problems__new_html_haml__3493347280847073492_69899255027720'
2015-05-18T11:45:12.629316+00:00 app[web.3]:   app/views/problems/_new.html.haml:13:in `_app_views_problems__new_html_haml__3493347280847073492_69899255027720'
2015-05-18T11:45:12.629318+00:00 app[web.3]:   app/views/layouts/application.html.haml:36:in `_app_views_layouts_application_html_haml___2941546639304470306_69899344274000'
```

justinfrench:
[...]

PR #1196, opened 2016-03-26
[rails] eager load formtastic

mikz:
should close https://github.com/justinfrench/formtastic/issues/911

### DIFF
--- a/lib/formtastic.rb
+++ b/lib/formtastic.rb
@@ -4,19 +4,22 @@ require 'formtastic/engine' if defined?(::Rails)
 module Formtastic
   extend ActiveSupport::Autoload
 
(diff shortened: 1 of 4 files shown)


mikz:
@justinfrench I think those constants should be the only ones that are not referenced on the class level and need to be eager loaded. The rest should be touched when the app is initialized. Wdyt? You know the internals better.

Eager loading all those constants will prevent from lazy loading them in production. Improving the memory footprint, speed and fixes concurrency issues by avoiding autoload on runtime.

justinfrench:
I haven't inspected this too closely, but looks okay and likely an improvement on where we are now :+1:

PR #1202, opened 2016-03-27
Change padding-left to margin-left for boolean labels

### LINKED ISSUE #1122: Use margin instead of padding for labels?
In formtastic.css, labels for booleans use `padding-left: 25%`. In Chrome, this makes the whitespace next to the checkbox clickable to toggle the checkbox, which seems somewhat user-unfriendly. Changing this to `margin-left: 25%` seems to fix the problem without ill effects.

Morred:
Fixes #1122


mikz:
Looks good! This will definitely need a CHANGELOG entry, as it can be a breaking change for many ppl.

Morred:
Do you want me to create one? If so, under which version number? (Latest entry stands at 3.1.2)

mikz:
@Morred I'll do it, no problem. cheers!

PR #1204, opened 2016-03-27
the changelog improvements

mikz:
it was quite hard to write a changelog for almost a year of changes
so I propose that we try to keep it in sync with what is merged

to do that, this pr adds missing entries and configures merge strategy
so conflicts are resolved by using both entries

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+CHANGELOG merge=union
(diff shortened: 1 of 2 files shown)


seanlinsley:
> to do that, this pr adds missing entries and configures merge strategy
so conflicts are resolved by using both entries

Does Github actually read `.gitattributes`? If so, a lot of OSS projects could benefit from that one-line change.

mikz:
@seanlinsley AFAIK no. But it is still useful when doing merges locally.

mikz:
@justinfrench merging this. IMO it will improve the workflow to push a new release.

PR #1205, opened 2016-03-27
formtastic integration test

mikz:
this creates an integration test to verify #1188 works

- [x] create new rails project
- [x] add formtastic
- [x] install formtastic
- [x] run tests
- [x] generate form
- [x] test form
- [x] improve bundle install speed on travis by reusing the same bundle path

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ Gemfile.lock
 doc/
 tmp
 gemfiles/*.lock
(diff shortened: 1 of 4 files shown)


mikz:
wdyt @justinfrench ? I think it is a nice way to verify the rails integration works correctly.

justinfrench:
Awesome idea!

PR #1160, opened 2015-07-13
Add symbols to raw_collection types that won't be mapped

Morred:
Attempt to fix justinfrench/formtastic#1127

PR #1206, opened 2016-03-27
upgrade rspec gem and syntax

mikz:
* upgrade to new syntax via [transpec](https://github.com/yujinakayama/transpec)
* upgrade rspec

@justinfrench wdyt? it is massive change done a by a tool

but imho it is useful as plenty of things in rspec are deprecated so it will be easier to upgrade in the future

all the suite is green, which is a good sign

the next step is to replace `rspec_tag_matchers` gem which causes more than 1k deprecations in the test suite


justinfrench:
Sounds great. My only worry is that this will instantly make most pull requests impossible to merge without work. There's probably no such thing as a _good time_ for this, but… Is it worth us cleaning up open PRs first?

mikz:
@justinfrench yep. There is just #1126 and #1160 that would be affected. Both are new tests, so should be easy enough to rebase them. I could do it if we decide to merge them.

justinfrench:
👍

PR #1208, opened 2016-03-27
Replace RSpec Tag Matchers

mikz:
`rspec_tag_matchers` was released in 2009 and haven't updated since.

This builds on #1206 and replaces the outdated gem with implementation based on rails `assert_select`. 
Also it fixes the `has_tag` calls with 2 parameters to 1 parameter and options hash. 

Depends on #1206.

I extracted the implementation into a new gem: https://github.com/mikz/rspec-dom-testing

### DIFF
--- a/formtastic.gemspec
+++ b/formtastic.gemspec
@@ -29,7 +29,7 @@ Gem::Specification.new do |s|
 
   s.add_development_dependency(%q<nokogiri>)
   s.add_development_dependency(%q<rspec-rails>, ["~> 3.4.2"])
-  s.add_development_dependency(%q<rspec_tag_matchers>, ["~> 1.0"])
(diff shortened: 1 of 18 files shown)


mikz:
@justinfrench wdyt? It is needed to drop the `rspec_tag_matchers` gem. There is `rspec-html-matches` replacement. But I find it not good enough. The same features can be done by just wrapping the assert_select from rails. I also plan to release this as a gem.

matthewrudy:
@mikz maybe should do this too to make the travis build clean.

mikz:
I'd like to but I was a bit worried that the whole testing does not work (because it was green too quickly with no edge cases) and does not work properly. So it requires a bit more of manual trying to break it and see it fails as it supposed to. And writing tests for the matchers of course.

PR #1209, opened 2016-03-28
Added Rubocop-coverage-coveralls.

ryanfox1985:
Added Rubocop.
Added Coveralls.
Fixed some offenses.

Can you register formtastic at https://coveralls.io? I Already added the badge.

Regards!


Gemfile
```diff
@@ -1,3 +1,9 @@
 source 'https://rubygems.org'
 
+group :development, :test do
+  gem 'simplecov', require: false
+  gem 'rubocop', '~> 0.37.2', require: false unless RUBY_VERSION =~ /^1.8/
```
mikz: I think we could use codeclimate to do the check instead of running it ourselves. Formtastic is already there. It is just matter of enabling rubocop check.

lib/formtastic/helpers/enum.rb
```diff
@@ -3,10 +3,8 @@ module Helpers
     # @private
     module Enum
       # Returns the enum (if defined) for the given method
-      def enum_for(method) # @private
-        if @object.respond_to?(:defined_enums)
-          @object.defined_enums[method.to_s]
-        end
+      def enum_for(method) # @private        
+        @object.defined_enums[method.to_s] if @object.respond_to?(:defined_enums)
```
mikz: I don't agree with the change. In this case, two line condition is great. Not only that you can actually measure if the branch is tested by coverage, but also it is more readable I'd say.

spec/inputs/custom_input_spec.rb
```diff
@@ -11,14 +11,16 @@ def input_args
     @method = :title
     @options = {}
     @proc = Proc.new {}
+
     if Rails::VERSION::MAJOR == 4
       @builder = Formtastic::FormBuilder.new(@object_name, @object, @template, @options)
-    else
-      @builder = Formtastic::FormBuilder.new(@object_name, @object, @template, @options, @proc)
+      return [@builder, @template, @object, @object_name, @method, @options]
```
mikz: this whole thing will be improved by removing the condition by https://github.com/justinfrench/formtastic/pull/1188/commits/b51d06d92f4f80560951ed8e5438a32c6f1fc487 (#1188)

spec/spec_helper.rb
```diff
@@ -1,3 +1,13 @@
+require 'coveralls'
+require 'simplecov'
```
mikz: this should be after `bundler/setup` because otherwise it can load different version than in the bundle (when executed without bundle exec).

mikz:
Good effort! I'd be definitely in for having a rubocop setting we agree on. 

However, I think we should use codeclimate for it, as it provides other checks too (like TODO,FIXME, ...) , can measure method complexity and find duplications and measure coverage. The project is already set up, but needs to be migrated to the new codeclimate platform. I'd need @justinfrench to do that/give me access. And of course, agree if it is the path he'd like to take.

Gemfile
justinfrench: @mikz if we only run on code climate, does that mean it's added as a build status thing alongside Travis?

justinfrench:
So, am I right that:

* There's a general consensus that Rubocop is a good idea regardless
* This PR adds Rubocop
* This PR fixes a bunch of issues caught by it (and a list of known issues that haven't not been fixed)
* Some of the fixes could be debated before this PR is merged

If we can focus this PR on the above, I think that would be great. 

Then there seems to be a choice about if this should be:

a) run in development and as part of the build on Travis
b) run on on CodeClimate
c) run on Coveralls

I'm no expert here, so if I'm being asked to decide, help me out be describing the advantages of each (and perhaps consider addressing it in a smaller follow-up PR). I'm more familiar with CodeClimate, so that's my default preference, although I'd love to hear from @ryanfox1985 on the merits of Coveralls over CodeClimate.

Finally, I'm kinda worried that `.rubocop_todo.yml` is _huge_ and I don't have a sense of where this is going. Would we leave current issues/offences as-is? Should we have a goal of reducing it? Should we have a goal of no issues? Would we need any contributor guidelines to help with this?

Gemfile
mikz: Yes. It would be another status check independent from Travis.

ryanfox1985:
Sorry but I not agree.

Code climate is a service focused in Ruby Guide Lines and best practices and Coveralls is a service for calculate the application coverage. There are different services with different objectives.

And I think it's necessary check the Rubocop before the push code to repository and not after, the same with the test. If you do after you are allowing bad smells to your repo.

I will investigate why is not passing the test.

Regards!

lib/formtastic/helpers/enum.rb
seanlinsley: unnecessary whitespace was added at the end of the the `def` line

mikz:
@ryanfox1985 

CodeClimate also measures coverage. It probably has less features than Coveralls (like failing the build if coverage drops). However, is quite nice to see both coverage and code quality at the same place.

You can totally run rubocop with codeclimate. The only difference is that codeclimate runs it on their infrastructure, instead of being a part of the main build. You still have your `.rubocop.yml` and all keeps working. 

There are things that codeclimate has as an extra (method complexity, duplication, todos), but they can be run via CLI (depends on docker). See http://blog.codeclimate.com/blog/2015/06/19/code-climate-platform/

I agree that there is a consensus that having a rubocop config is a good idea. Having rubocop as part of the build is better than to not have it at all.

I don't agree with failing travis build because of style violations. Not only because in my opinion those two don't mix (as they build status can be enforced or not). But also because it is quite hard to extract what is wrong from the travis output. 

If we would be talking about enforcing style guide, something like [hound](https://houndci.com/) might be worth looking for, just because it comments on the PR (both blessing and a curse).

Having CodeClimate to do method complexity analysis, find duplicated code, etc. is a good thing in my opinion. And batching it with rubocop makes sense to me, because they are similar thing and belong together to a different status check than the main build. 

just my 2ç 

But still, having this is better having this :)

seanlinsley:
IRT Hound, you don't want that until you have a consistent style. In files with existing violations it will, well, hound PR authors even though those violations weren't added by them.

ryanfox1985:
Hund is using Rubocop https://houndci.com/configuration#ruby

mikz:
@ryanfox1985 this PR is not mergeable now. Looks like the autocorrect conflicts with basically anything merged. Maybe would be better to leave it out and do it as a separate PR ?  With it is is kind of impossible to review.

ryanfox1985:
I removed the two last commits.

ryanfox1985:
Can @justinfrench add the repository in coveralls.io?

Gemfile
mikz: Why is there Ruby version check? There is no support for Ruby 1.8. That should not be there at all.

lib/formtastic/helpers/enum.rb
mikz: @ryanfox1985 would be good to agree on a code style before changing it. I really don't find this more readable.

spec/spec_helper.rb
mikz: @ryanfox1985 this is still valid comment.

spec/spec_helper.rb
```diff
@@ -17,6 +27,10 @@ module VERSION
   end
 end
 
+require 'rubocop/rake_task'
+RuboCop::RakeTask.new
+Rake::Task['rubocop'].invoke
```
mikz: Not sure if that is a good idea. If we want to run rubocop, it should be just another line in travis.yml. It should not be executed when running specs.

justinfrench:
Hi folks,

I'm finding it difficult to follow this thread and nerd-out on so many details to make a good decision. I'm going to defer to @mikz on any final decisions, and he can let me know if/when new systems need to be set-up/connected.

However, here's a few preferences of mine from me that might help guide debates:

1. Unless someone is volunteering to actively improve the violations found by rubocop, I don't actually see the point. In other words, if we end up _not merging this_, that's okay with me. Same goes for Code Climate, Coveralls, etc. It's just noise and a barrier to contribution unless someone is motivated to improve things. At the moment, that's not me.

2. At the moment, the only code tests we have are specs. We invoke them with `rake spec` on CI. If we add any new kinds of code tests (for coverage, or code style, or whatever), I don't want them bolted onto `rake spec`. We should add a new task `rake ci` which subsequently calls `rake spec` and other new rake tasks for new codes of tests. This way I can choose to…

  * run just the specs, or just the coverage, or just the code style stuff
  * run everything, just like it will be run on CI

I'll leave this one with @mikz to decide how to best proceed.

PR #1210, opened 2016-03-28
drop support for ruby 1.9.3

mikz:
EOL was more than a year ago
https://www.ruby-lang.org/en/news/2014/01/10/ruby-1-9-3-will-end-on-2015/

the travis build is failing anyway, because it can't install some gems (requiring >= 2.0)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,6 @@ cache:
     - vendor/bundle
 language: ruby
 rvm:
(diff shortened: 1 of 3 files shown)


mikz:
@justinfrench I'd like your +1. This is pretty significant bump.

justinfrench:
Yep :+1:, thanks

Issue #1187, opened 2016-01-22
Password field inputs are shown as regular inputs in Active Admin


arogachev:
Initially I described problem here https://github.com/activeadmin/activeadmin/issues/4280 and thought that it was Active Admin problem.

I'm using Rails `5.0.0.beta1`, My `Gemfile` contents:

```ruby
gem 'activeadmin', github: 'activeadmin/activeadmin', branch: 'rails-5-rspec'
```
I tried to update to latest commit of Formtastic on master branch, that didn't help:

```ruby
gem 'formtastic', github: 'justinfrench/formtastic', branch: 'master'
```

For example, this line from [this file](https://github.com/activeadmin/activeadmin/blob/master/app/views/active_admin/devise/sessions/new.html.erb):

```ruby
f.input :password, label: t('active_admin.devise.password.title')
```
generates the following HTML:

```html
<input type="text" name="user[password]" id="user_password" maxlength="72">
```

So there is no `type="password"`.

mikz:
@arogachev does this happen for all the fields or just devise ?
This might be devise problem. https://github.com/plataformatec/devise/issues/3880

arogachev:
@mikz Thanks for fast reply. If i write something like this inside resource page:

```ruby
form multipart: true do |f|
  f.inputs do
    f.input :password, as: :password
  end
end
```

input is shown correctly.

I can't fully understand magic behind:

f.input :password, label: t('active_admin.devise.password.title')

Here the type is not explicitly specified, so password is rendered by name `password`?

mikz:
So the change that broke this is https://github.com/jtomaszewski/formtastic/commit/79b86c5ee0becb1cd2a795e3883c5bc319caaf83
For devise inputs, the column has no type, so we just default to string. However, we could default the type to string and check for the method name. Will fix this asap.

mikz:
@arogachev could you either provide sample project, or try to debug [the code in question](https://github.com/jtomaszewski/formtastic/blob/88f28a0bc2cdea61347b5f773ce058d68ce2fc4f/lib/formtastic/helpers/input_helper.rb#L257-L299)?

I see that it [should return `:password`](https://github.com/jtomaszewski/formtastic/blob/88f28a0bc2cdea61347b5f773ce058d68ce2fc4f/lib/formtastic/helpers/input_helper.rb#L296 even if the column type is nil) (devise case).

[...]

Issue #1143, opened 2015-03-21
Problem with color_field form helper 


andywenk:
Hi,

I saw that you integrated support input[type=color] fields in https://github.com/justinfrench/formtastic/issues/419

I stumbled over this problem:

    [ERROR]: RuntimeError (The :color input requires the color_field form helper, which is only available in Rails 4+):
      /var/www/www.qraex.de/shared/bundle/ruby/2.2.0/gems/formtastic-3.1.3/lib/formtastic/inputs/color_input.rb:34:in `to_html'
      /var/www/www.qraex.de/shared/bundle/ruby/2.2.0/gems/formtastic-3.1.3/lib/formtastic/helpers/input_helper.rb:240:in `input'
      /var/www/www.qraex.de/shared/bundle/ruby/2.2.0/gems/arbre-1.0.3/lib/arbre/rails/forms.rb:30:in `proxy_call_to_form'
      /var/www/www.qraex.de/shared/bundle/ruby/2.2.0/bundler/gems/activeadmin-95127afc924d/lib/active_admin/views/components/active_admin_form.rb:63:in `input'
      /var/www/www.qraex.de/releases/20150213144943/app/admin/cartridges.rb:44:in `block (3 levels) in <top (required)>'

I have some fields in the database like this:

    f.inputs "Cartridge erstellen" do
    f.input :name
    f.input :qraexid
    f.input :oem
    f.input :capacity
    f.input :q_capacity
    f.input :pages
    f.input :photo_pages
    f.input :bottle
    f.input :cap_sign
    f.input :cap_color
    f.input :cap_color_help
    f.input :mode
    f.input :common_price
    end

The error is pointing to `f.input :cap_color`. Am I required to change the name? As far as I understood the code, it should only be a `type=color` field when writing `f.input :cap_color, as: :color `. Any idea what is wrong here?

Thanks a lot and cheers

Andy

justinfrench:
Hi, I'll take a look at the code later, but the workaround is to specify an explicit `:as => :string` (or similar) to bypass any automatic choices in the code.

andywenk:
Thanks a lot Justin!

mikz:
Right now it chooses :color type when the column contains `color`. https://github.com/justinfrench/formtastic/blob/2f8b987ee4880c4aa06e6ec799eafe48fa46e316/lib/formtastic/helpers/input_helper.rb#L273.
We dropped support for Rails 3. So `color_field` should be available everywhere.

[...]

PR #1211, opened 2016-04-03
[experiment] Configurable mapping of columns to inputs

mikz:
This is a proof of concept to offer ability to configure column to input mapping per form builder.

Some people want to map a column to different input type (#1165, #1143). Some people want a default (#1183). Some would like to have an override (#1187).

proposed in https://github.com/justinfrench/formtastic/issues/1146#issuecomment-203114072

I'd like to get some comments on this approach and verify it solves problems raised in #1146.
I'm not saying I'm happy with this particular API, it is too early for that. But IMO the way is right.

/cc @justinfrench @timoschilling @jclay
## Examples

I have not tried them. But they should work. The idea is to verify the API is ok for those use cases and then actually implement it and have a test coverage for those.
### Adding inet (https://github.com/justinfrench/formtastic/issues/1146#issuecomment-127635091)

``` ruby
string_input = Formtastic::FormBuilder.input_mapping.add_input(:string)
string_input.match_type(:inet)
```
### Mapping Hstore to text (#1165)

``` ruby
text_input = Formtastic::FormBuilder.input_mapping.add_input(:text)
text_input.match_type(:hstore)
```
### Mapping color to string (#1143) (not implemented yet)

``` ruby
string_input = Formtastic::FormBuilder.input_mapping.prepend_input(:string)
string_input.match_column(/color/)
```
### Having global default (#1183) (should change)

```
string_input = Formtastic::FormBuilder.input_mapping.add_input(:string)
string_input.match_type(nil)
```
### Having a name without a type (#1187)

```
password_input = Formtastic::FormBuilder.input_mapping.add_input(:password)
password_input.match_type(nil)
password_input.match_column(/password/)
```
### TODO
- [ ] figure out better way to handle defaults than `Input#match_type(nil)`
- [ ] on form builder inheritance, the input mapping gets copied and possibly mutated

### DIFF
--- a/lib/formtastic/form_builder.rb
+++ b/lib/formtastic/form_builder.rb
@@ -47,6 +47,155 @@ module Formtastic
     configure :action_namespaces, [::Object, ::Formtastic::Actions]
     configure :action_class_finder, Formtastic::ActionClassFinder
 
+    class InputMapping
+
+      class Input
+        attr_reader :name
+
+        def initialize(name)
+          @name = name
+          @matchers = Hash.new { |h,k| h[k] = [] }
+        end
+
+        def match_column(name)
+          @matchers[:column] << name
+        end
+
+        def match_type(type)
+          @matchers[:type] << type
+        end
+
+        def match_form(matcher = nil, &block)
+          @matchers[:form] << (matcher || block)
+        end
+
+        def matches_type?(type)
+          match?(:type, type)
+        end
+
+        def matches_column?(name)
+          match?(:column, name)
+        end
+
+        def matches_form?(*args)
+          match?(:form, args)
+        end
+
+        protected
+
+        def match?(name, match)
+          matchers = @matchers[name]
+          matchers if matchers.none? || matchers.any? { |matcher| matcher === match }
+        end
+      end
+
+      def initialize(mappings = [])
+        @mappings = mappings
+      end
+
+      def add_input(name)
+        @mappings << input = Input.new(name)
+        input
+      end
+
+      def to_a
+        @mappings.to_a
+      end
+
+      def input_name
+        input = @mappings.first
+        input && input.name
+      end
+
+      def select_type(column)
+        type = column && column.type
+
+        InputMapping.new @mappings.select { |m| m.matches_type?(type) }
+      end
+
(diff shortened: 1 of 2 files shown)


mikz:
Ping @justinfrench @timoschilling @jclay

No rush. Thaler your time. Would be cool to know if you feel it is right approach or not.

justinfrench:
Hi @mikz, this is a really cool experiment, but I don't _think_ I like it. There's lots to discuss, but perhaps I'll start with two small example.

```
select = input_mapping.add_input(:select)
select.match_type(:integer)
select.match_form { |form, method| form.reflection_for(method) }
select.match_form { |form, method| form.enum_for(method) }

# ...

number = input_mapping.add_input(:number)
number.match_type(:integer)
number.match_type(:float)
number.match_type(:decimal)
```

I have two main observations:

The first is that I can see two inputs being added/registered, both of which `match_type(:integer)`, but it's not clear to me which one has precedence. Is it the first one added? The last one? I agree that the way we currently do it isn't _configurable_, but at least it's much clearer which rules take precedence.

The second is that It's not really clear to me if the `match_*` rules are `AND` or `OR`. I _think_ they're an `OR` — For the Number input you're saying "Register a Number input, use it for Integer or Float or Decimal columns".

That's pretty clear, but for the Select input it gets a bit trickier: "Register a Select input, and use if for Integer columns or when there's a AR reflection or when there's an Enum". Like I said above, we have two different inputs for Integer columns, but the code we currently have in master isn't an `OR`. We have a combination of `AND` and `OR`:

```
when :integer
  return :select    if reflection_for(method)
  return :select    if enum_for(method)
  return :number
```

* Use a select input if the column is an integer AND (there's a reflection OR there's an enum)
* Otherwise, use a number input if the column is an Integer

In short, I think this API makes it far less clear what's happening in a pretty complex decision tree that we've evolved over a long time. I'm not 100% sure what "better" would look like though!

mikz:
@justinfrench awesome feedback! I have no idea how the ideal interface would look like, but maybe we can get there with some experimentation :) Will give it some time and try to come up with another experiment.

justinfrench:
Closing for now due to inactivity

Issue #1127, opened 2015-02-09
Symbols support is incomplete


romanbsd:
While using with the role_model gem, the following inconsistency has surfaced:
- The user.roles returns a Set of Symbols
- The User.valid_roles is an array of Symbols

While the selected_values (make_selected_values) is an array of Symbols, the Collections#collection converts the User.valid_roles to array of strings because

``` ruby
[Array, Fixnum, String].include?(raw_collection.first.class) #-> false
```

As a workaround, I provide the collection in the form:

``` ruby
User.valid_roles.map { |role| role.to_s.humanize }.zip(User.valid_roles)
```

justinfrench:
Sounds like we can treat symbols like strings. Would love to see a patch!

romanbsd:
I'm not familiar with the formtastic internals, thus I don't have a slightest idea what can break if the Symbol is added to the list above.

Morred:
I gave this a try, so now arrays of symbols won't be mapped anymore but will be returned in their original form just like an array of strings would be. If that's not what you were aiming for here, I'm open for suggestions :)

justinfrench:
@Morred @romanbsd could one of you confirm that the #1160 patch fixes this issue?

Morred:
In combination with the gem @romanbsd was mentioning?

justinfrench:
@Morred yeah, the fix _looks_ good, but I'd love to know for sure this fixes the problem it attempts to solve before merging.

Morred:
@justinfrench Absolutely, that makes sense. I'd love for @romanbsd to try it out because he could definitely tell whether it solves the problem in his current app or not. If he doesn't have time, I'll set up a small test app and check it myself.

justinfrench:
@romanbsd? @Morred?

Morred:
Ah damn, completely forgot about this one! If @romanbsd isn't here anymore, I guess I'll have to be the one to try it out

PR #1216, opened 2016-05-24
Apply symbol value like string

kvokka:
Apply symbols for checkboxes, like strings. It makes interface more clear, and prevent from some night errors :)

### DIFF
--- a/lib/formtastic/inputs/check_boxes_input.rb
+++ b/lib/formtastic/inputs/check_boxes_input.rb
@@ -138,11 +138,11 @@ module Formtastic
       end
 
       def checked?(value)
-        selected_values.include?(value)
+        selected_values.include?(value.is_a?(Symbol) ? value.to_s : value)
       end
 
       def disabled?(value)
-        disabled_values.include?(value)
+        disabled_values.include?(value.is_a?(Symbol) ? value.to_s : value)
       end
 
       def selected_values


mikz:
Is this related to #1127 ?

To merge this, I'd need tests. But still I'm not sure if it is the right way. Models are free to return symbols or strings as they please and it should be responsibility of the programmer to keep using the same.

This could lead to some situations when user would not be able to pass in symbol even when really needed.

kvokka:
I did not see this issue report, but spent some time for understanding, what is wrong
I understand, that it is possible to return symbol by the model, but, actually, in this case, I think it does not matter. I've done online patch (It was simpler), I'll look in tests soon, so it'll pass soon.
Of course, if this patch is good enough

mikz:
What matters is case when:

* model returns a symbol
* the collection items are symbols

Then you'd cast that symbol from model to a string and break it.

The problem is: what if I use symbols everywhere? I have an app that does in some cases. This would break it.

kvokka:
Good argument, but I can provide both checks, so it will not matter, is it String or Symbol. Of course :foo == 'foo' in this case. But, as for me, for checkboxes it is much closer for principle of least astonishment than current way.

mikz:
Providing both would probably make sense.
But then maybe it would be just fine to call `selected.include?(object) || selected.include?(object.to_s)` without any is_a?.  But I'm really not sure there.

kvokka:
Yes, your solution will be faster. So, I think it will be good to implement yours.

kvokka:
Moved to https://github.com/justinfrench/formtastic/pull/1217
Remade code, tests pass

justinfrench:
Closing as there's a newer PR

justinfrench:
I _think_ #1160 covers this _and`_ includes test coverage, so I'm going to prefer that one over this one and #1217 for now.

PR #1219, opened 2016-05-25
Check boxes control is not saved as empty array if no boxes were checked (master)

sandfoxme:
Undo merge #1190, there was no bug there

The issue #1189 however stands valid: `hidden_fields?` seems poorly named. Maybe rename it `hidden_fields_for_every?` to oppose `hidden_field_for_all`

### DIFF
--- a/lib/formtastic/inputs/check_boxes_input.rb
+++ b/lib/formtastic/inputs/check_boxes_input.rb
@@ -99,7 +99,7 @@ module Formtastic
       end
 
       def hidden_field_for_all
-        if hidden_fields?
+        if hidden_fields_for_every?
           ''
         else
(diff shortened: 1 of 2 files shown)


sandfoxme:
Oh, this is a fix for #1219

mikz:
Thanks! It makes total sense. 

That naming is terrible and has to be fixed. 4.0 would be a good time to do it. 

Would you care to take a stab in it and try to improve it a bit?

Btw, the tests fail.

sandfoxme:
whoops. forgot to add expect() around output buffer

* renamed the method
* tests are passing now

mikz:
Great. I think now I finally understood it :)

PR #1227, opened 2016-08-17
Fix i18n for enums inside nested forms

### LINKED ISSUE #1151: collection_from_enum I18n unusable with has_many nested forms
the `collection_from_enum` method try to fetch translation with the following key: 

``` ruby
key = "activerecord.attributes.#{object_name}.#{method_name.pluralize}.#{name}"
```

In case where the field is in a has_many nested (eg: with cocoon) form, object_name can be like this : 

``` ruby
car[wheels_attributes][0]
```

So it is impossible to use the translation.
I suggest to change object_name by `model_name.underscore` or something like this to obtain just the model name as expected to use in the translation yml file.

arokettu:
This fixes #1151

### DIFF
--- a/lib/formtastic/inputs/base/collections.rb
+++ b/lib/formtastic/inputs/base/collections.rb
@@ -110,12 +110,28 @@ module Formtastic
         #           archived: Custom Archived Label Here
         def collection_from_enum
           if collection_from_enum?
-            method_name = method.to_s
+            method_name       = method.to_s
+            object_class      = object.class
+            lookup_ancestors  = object_class.lookup_ancestors
+            attributes_scope  = "#{object_class.i18n_scope}.attributes"
+
+            # apply i18n discovery algorithm close enough to algorithm from
+            # ActiveModel::Translation.human_attribute_name
+            # activemodel-5.0.0.1:lib/active_model/translation.rb:43
 
             enum_options_hash = object.defined_enums[method_name]
-            enum_options_hash.map do |name, value|
-              key = "activerecord.attributes.#{object_name}.#{method_name.pluralize}.#{name}"
-              label = ::I18n.translate(key, :default => name.humanize)
+            enum_options_hash.map do |name, _value|
+              attribute = "#{method_name.pluralize}.#{name}"
+
+              defaults = lookup_ancestors.map do |klass|
+                :"#{attributes_scope}.#{klass.model_name.i18n_key}.#{attribute}"
+              end
+
+              defaults << :"attributes.#{attribute}"
+              defaults << name.humanize
+
+              options = { defaults: defaults }
+              label = ::I18n.translate(defaults.shift, options)
               [label, name]
             end
           end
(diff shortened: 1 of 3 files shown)


sandfoxme:
The idea is that `key = "activerecord.attributes.#{object_name}.#{method_name}"` is not really a correct way to get activerecord i18n data, there is a discovery algorithm that also looks for object ancestors

The algorithm is a part of ActiveModel

here is the method itself taken from 

```
    def human_attribute_name(attribute, options = {})
      options   = { count: 1 }.merge!(options)
      parts     = attribute.to_s.split(".")
      attribute = parts.pop
      namespace = parts.join("/") unless parts.empty?
      attributes_scope = "#{self.i18n_scope}.attributes"

      if namespace
        defaults = lookup_ancestors.map do |klass|
          :"#{attributes_scope}.#{klass.model_name.i18n_key}/#{namespace}.#{attribute}"
        end
        defaults << :"#{attributes_scope}.#{namespace}.#{attribute}"
      else
        defaults = lookup_ancestors.map do |klass|
          :"#{attributes_scope}.#{klass.model_name.i18n_key}.#{attribute}"
        end
      end

      defaults << :"attributes.#{attribute}"
      defaults << options.delete(:default) if options[:default]
      defaults << attribute.humanize

      options[:default] = defaults
      I18n.translate(defaults.shift, options)
    end
```

sandfoxme:
The fix is not final, it fails some tests. But I hope it is the direction in which we should look

sandfoxme:
* fixed one bug with i18n fallbacks
* classes in tests now also mock ActiveModel::Name and lookup_ancestors
* added a test for ancestor translation lookup

tests are passing now. seems final enough

sandfoxme:
Reverted some changes in tests: Post already uses ActiveModel::Naming and MongoPost has nothing to do with the task

avaz:
Hey guys,

Any idea when this will be available?

Thanks!

Nghi93:
any updates on this?

mikz:
We could aim to have this in 4.0. Would it help the situation to support only Rails 5+?
I don't really like the fact that we would copy a rails method inside formtastic, so it would be better to just reuse what is there in Rails.

Also, I'd like to have a clean git history, so the whitespace changes should be applied in different PR for easier review.

sandfoxme:
Could anyone take it from me? I don't even work with ruby currently

Fs00:
Hello everyone. If the maintainers (@justinfrench @deivid-rodriguez) are still interested in merging a fix, I would like to propose a simpler solution to this issue in another PR.
I've stumbled upon it earlier today and has been pretty annoying.

mikz:
@Fs00 I think that would be awesome! Would be awesome if it could use Rails API as much as possible, so we don't have copy a method implementation. And I guess and explanation what the problem is with some examples would be awesome :) Thank you for doing this!

Fs00:
@mikz This can now be closed since #1342 has been merged :)

mikz:
Fixed by #1342

PR #1203, opened 2016-03-27
Added Hstore type.

ryanfox1985:
Ref: #1165.

I removed the HstoreInput class that I copied from TextInput and added hstore type at the default_input.

Issue #1197, opened 2016-03-26
Formtastic 4.0


mikz:
New release of Formtastic is coming. https://github.com/justinfrench/formtastic/pull/1108#issuecomment-187905777

Major changes will be:
- [x] dropping support for Rails 3 and Rails 4.0  (#1108)
- [x] support for Rails 5 Attributes API (#1188)
- [x] eager load to prevent errors in concurrent environments (#1196)
- [x] removal of deprecated finder methods (#1139)
# TODO
- [x] release 3.1.4 #1198
- [x] merge #1108
- [x] merge #1139 
- [x] merge #1196
- [x] merge `4.0-dev` to master
- [x] switch 4.0 work to master
- [x] reevaluate [4.0 beta milestone](https://github.com/justinfrench/formtastic/milestone/27) 
- [x] add more tasks
- [x] release 4.0 beta
- [ ] release 4.0 rc
- [ ] release 4.0

voyera:
Just have to say keep up the good work guys, this is great!

javierjulio:
@mikz in your first todo list, it seems that "support for Rails 5 Attributes API (#1188)" should be marked completed since PR was merged. Is a v4 release of Formtastic still planned? Anyone pushing that through or has effort died out?

mikz:
@javierjulio thanks for checking. Marked it as done. 

It was just me pushing through and I could not find the time lately. I guess 

What would help is someone trying the git branch and checking if there are any issues with the upgrade. Then we could mark the beta release I guess.

javierjulio:
@mikz I'm asking since myself and a few others are new maintainers of ActiveAdmin. Formtastic gem is pretty important to us so we'd like to see it continue or look for a new solution. It seems the work is here it just needs some attention and a release. One of the last PR's you took was from @Fivell who is also an ActiveAdmin maintainer.

I set Formtastic to master branch on AA locally, ran our test suite and everything is green. 👍🏻 I will do more testing with our sample app which has various form setups and that should really test Formtastic v4. How else can we help to get a release out? 

While its possible to modify the wrapping HTML of inputs, the fieldset/list wrapper is a problem. Would you also consider a change to make it easier to override the fieldset and list wrapper HTML output? Right now its in a module but done in a long method without a way of doing it. It would be helpful to have a way of modifying the HTML for all in v4 even if it means subclassing all inputs.

mikz:
[...]

PR #1229, opened 2016-09-22
Default input type should be text for json, jsonb, citext

swrobel:
Currently these just throw an exception like "No input found for <type>." This is in the same spirit as https://github.com/justinfrench/formtastic/pull/1203, which is for `hstore`, a similar type to `json`. `citext` is a simply case-insensitive version of `text`.

### DIFF
--- a/lib/formtastic/helpers/input_helper.rb
+++ b/lib/formtastic/helpers/input_helper.rb
@@ -283,8 +283,10 @@ module Formtastic
             return :time_select
           when :date
             return :date_select
-          when :hstore
+          when :hstore, :json, :jsonb
             return :text
+          when :citext
+            return :string
           end
 
           # Try look for hints in options hash. Quite common senario: Enum keys stored as string in the database.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
   * Added mapping hstore column to text input (#1203)
   * Added support for Rails 5 Attributes API (#1188)
   * Changed required Ruby version to >= 2.0 (#1210)
+  * Default to input types text for json & jsonb, string for citext columns (#1229)
 
 ## 3.1.2
 
--- a/spec/helpers/input_helper_spec.rb
+++ b/spec/helpers/input_helper_spec.rb
@@ -482,8 +482,10 @@ RSpec.describe 'with input class finder' do
           expect(default_input_type(:date)).to eq(:date_select)
         end
 
-        it 'should default to :text for :hstore column types' do
+        it 'should default to :text for :hstore, :json and :jsonb column types' do
           expect(default_input_type(:hstore)).to eq(:text)
+          expect(default_input_type(:json)).to eq(:text)
+          expect(default_input_type(:jsonb)).to eq(:text)
         end
 
         it 'should default to :datetime_select for :datetime and :timestamp column types' do
@@ -501,6 +503,10 @@ RSpec.describe 'with input class finder' do
 
         it 'should default to :string for :string column types' do
           expect(default_input_type(:string)).to eq(:string)
+        end
+
+        it 'should default to :string for :citext column types' do
+          expect(default_input_type(:citext)).to eq(:string)
         end
 
         it 'should default to :number for :integer, :float and :decimal column types' do


mikz:
Looks good. Could you add a test and a changelog entry? Then I could merge it right away.

frodsan:
@swrobel @mikz any updates on this? I could add what's missing.

mikz:
I'll not be available for next week to review this. Do you mind pinging me later?

frodsan:
ok 👍

swrobel:
I'll add a changelog entry. @mikz not sure what you want in terms of a test ... I piggybacked off the one that already existed for `hstore` ... not sure what else would be needed.

lib/formtastic/helpers/input_helper.rb
```diff
@@ -283,7 +283,7 @@ def default_input_type(method, options = {}) # @private
             return :time_select
           when :date
             return :date_select
-          when :hstore
+          when :hstore, :json, :jsonb, :citext
             return :text
```
swrobel: text database type renders a text input, so it makes sense to me that citext would also render text.

mikz:
That looks fine for me. If @frodsan agrees that text input is fine, I'll be happy to merge this.

frodsan:
I think it must be an input text field instead of textarea:

When `:text` is returned, it looks like this in ActiveAdmin:

screen shot 2016-12-05 at 16 26 25

I understand why you would like a textarea for Hstore or JSON because there is probably a lot of content, but citext is only used to avoid this pattern:

```
WHERE LOWER(username) = LOWER('My@maiL.com')
```

With bigger text, I'd use PG full text search.

https://www.postgresql.org/docs/9.6/static/citext.html

> The citext module provides a case-insensitive character **string** type, citext.

frodsan:
@swrobel It would be nice to know what's is your use case, like why do you need to edit a citext attribute in a textarea?

mikz:
Agree that citext is different from hstore/json and should default to a string.

But then there is not much point for adding it as the default type is string anyway.

With the rails 5 attribute it should be possible to control this on the model (https://github.com/justinfrench/formtastic/commit/1fb2e2490f2a69384d765bfd6065f40cd5ebe5fb).

swrobel:
@frodsan you make a good point. I use it for email addresses as well and will update the PR.

frodsan:
I get the following error when I don't use this patch:

```
Formtastic::UnknownInputError at ...
Unable to find input class CitextInput
```

Is the new attributes API used for custom attributes? Citext is a type supported by default in Rails (https://github.com/rails/rails/blob/b3e0da3062f809599fb4da9f9a29ed45aebd1ff6/activerecord/lib/active_record/connection_adapters/postgresql/oid.rb#L376..L379)

mikz:
@frodsan [the test](https://github.com/justinfrench/formtastic/blob/06c5d47a7e4a64b61fa53a39adf22fa1f9681716/script/integration/rails-5-0.rb) overrides string to password type. Imo it can be used for both new types and existing ones.

frodsan:
@mikz We've added this to the project:

```
# frozen_string_literal: true
module Formtastic
  module Inputs
    class CitextInput < StringInput
    end
  end
end
```

mikz:
@frodsan looks like it tries to load `CitextInput` from the global scope.

The default says it should try to load from both. https://github.com/justinfrench/formtastic/blob/06c5d47a7e4a64b61fa53a39adf22fa1f9681716/lib/formtastic/form_builder.rb#L43-L45


https://github.com/justinfrench/formtastic/blob/06c5d47a7e4a64b61fa53a39adf22fa1f9681716/lib/formtastic/namespaced_class_finder.rb is responsible for finding the right class. Maybe you can dig there.

But this is not really related to this PR or Rails 5 attributes API, right?

frodsan:
@mikz Sorry for short explanation. It's not related to this PR or the Rails 5 attribute API, but it's something we did to fix our case without waiting for a next version. It's my first time using formtastic.

mikz:
@frodsan no worries. Going to merge this ❤️ PR.

Regarding using custom input types in formtastic, there should be a way to set up custom namespace for them and control how they are found. The feature got introduced in https://github.com/justinfrench/formtastic/pull/1063

swrobel:
@mikz any reason this wasn't released with 3.1.5?

mikz:
@swrobel 3.1.5 had just one change - support Ruby 2.4. I'd say changes like this should go to minor version releases like N.X rather than patch releases. There is ongoing effort to ship 4.0 (https://github.com/justinfrench/formtastic/issues/1197). Would that work for you?

swrobel:
@mikz yep, that works!

PR #1230, opened 2016-09-22
Convert CHANGELOG to Markdown

swrobel:
Having an ambiguous CHANGELOG format breaks parsers like [vandamme](https://tech-angels.github.io/vandamme/).


mikz:
👍

PR #1232, opened 2016-11-18
Fix documentation link in README

rglassett:
The current documentation link fails with a 503 backend fetch failed error. If there's a solution on the Rubydoc side that might be preferable, otherwise this seems to do the trick.


mikz:
Loads just fine for me. Redirects to `http://www.rubydoc.info/github/justinfrench/formtastic`. Which might be better choice.

rglassett:
Confirmed working for me as well. Must have been a temporary issue, I'll close this PR.

PR #1238, opened 2017-01-24
Add a config to allow overriding collections retrieval methods

Uelb:


mikz:
Could you expand a bit on the use case and why it should be in formtastic itself?

Uelb:
Hello @mikz,
Thank you for your answer.

I'm not sure it should be in formtastic, but I found no way to actually cover this case for my project and I did it like that. I'm sorry I had no time to explain more yesterday.

This feature is useful when you need to override how collections are retrieved by formtastic, let's say you have a Post model : 
````
class Post
  has_one :company
  belongs_to :user
end
````

a User model 
````
class User
  has_many :posts
  belongs_to :company
end
````

a Company model
````
class Company
  has_many :users
  has_many :posts, through: :users
end
````

You may be designing a form to allow a user to add a post for anybody else in his company.
There is today a way to do that using formtastic by directly passing the collection to the input of course:

````
# Simplified
form do |f|
  f.input :user, collection: proc {User.where(company: current_user.company)}
end
````

But there is no way of implementing a default behavior for this retrieval. In my case I'm using active admin with cancancan for authorization, and as formtastic handles automatically the retrieval, even if I do not want one of my user to see some associated records, he sees them all unless I specify the collection to every single input of every form of the app.
I also thought it would be a nice way to allow gems depending on formtastic (activeadmin included) to actually be able to override that part of the form. I may write another PR on activeadmin later to be able to scope the filters and form inputs of active admin to be scoped to the AuthorizationAdapter they let the user create.

For my case, I added that in my initializer
````
# config/initializers/formtastic.rb
Formtastic::FormBuilder.base_input_collections_module_override = CustomBaseForActiveAdmin
````

And I overrided the collection_from_association method as I wanted to.
I first tried to let active admin provide this level of customization but I could not do much as every inputs includes a module and not a class, so it's difficult to reopen it and just edit methods, although I may have missed something.

Of course, I'm not qualified enough to know if this should be inside Formtastic or somewhere else, it also depends on the level of customization you want to provide and I'll let you judge if you think it's a nice feature to have.

**Comment by mikz** [comment]
@Uelb I see. So for overriding `collection_from_association` in some global way to limit scope of the records returned.

Could you share some of those customisations to see what kind of collaborators it needs?
I'd like to see if that requires just the model or also view/controller/form. 

Maybe the solution would be just to specify a scope option.

**Comment by Uelb** [comment]
Sorry for the late answer.
Here is the (simplified) code I use so that every resources scopable to a company is scoped automatically and I ensure there is no issue on what the user is able to see.
````ruby
module CustomBaseForActiveAdmin
  def self.included(base)
    base.send :alias_method, :old_collection_from_association, :collection_from_association
  end

  def collection_from_association
    if reflection
      if object.class != Ransack::Search && reflection.klass.respond_to?(:with_company)
        agency_id = if object.respond_to?(:agency)
          if object.persisted?
            object.agency.id
          else
            template.current_user.agency_ids
          end
        end
        return super unless agency_id
        super.with_agency(agency_id)
      else
        super
      end
    end
  end
end
````

**Comment by mikz** [comment]
What about using a Symbol for the collection option like:

```ruby
f.input :user, collection: :company_users
```

Or just using a proc:

```ruby
f.input :user, collection: &:company_users
```

And implementing a helper method.

I think using code in all collection inputs does not help readability or clarity in the form definitions and there are better means to achieve the same. 

I'd consider implementing `FormBuilder#collection_for_reflection(reflection)` or even better abstracting it into own class like `Formtastic::InputClassFinder` or `Formtastic::ActionClassFinder` and having this object just single responsibility to find a relation for a reflection and some form object.

Uelb:
I see 2 potential issues with your approach : 
- There is no way to be sure that every input has the wanted behavior, and thus a company would if omitted be able to see other users company.
- One has to repeat the collection: ... for every input, and that's a lot of repetition when using formtastic a lot

I'm not sure I understand how this way would be different from today as we can already use a proc and access the current_user for instance inside the proc. The proc can also be a variable so it would work as a helper method.
The point of the pull-request is not to have to do that for every input when you want to implement a global behavior for your form or you have critical data segregation to do.

Finally, I have not understood your last sentence, do you mean that you would consider putting a collection_for_reflection method in another class used in the FormBuilder module ? In that case, it would be easy for any user to extend this class and I guess it would solve the use case. The main issue I had is that most of the code is written in modules so it's more difficult to extend it in "custom code" than if it was a class

mikz:
Agree that having it in every input can be tedious and can be error prone, but that is why we have tests.

Methods are easier testable than proc in view. Imo there is no need for proc in view (i'd even try to use render collection instead of .each). 

But to stick to the point of having a global way to modify how collections are found. I'd consider putting it on the FormBuilder module or even better extracting it into a class and allow configuration like: https://github.com/justinfrench/formtastic/blob/ec0b090de764f09cd7d68177ac6fea9011d907b4/lib/formtastic/form_builder.rb#L48
That is later used: https://github.com/justinfrench/formtastic/blob/b104bb1fe5228ea37f668917b13d0888f8bd033a/lib/formtastic/helpers/action_helper.rb#L107-L109

Then you'd just be able to configure the collection finder globally or per form.


And regarding this PR, I think it can be easily just a monkey patch in your application using `Module#prepend`:

```ruby
module MyGlobalModule
  def collection_from_association
    if reflection
      User.where(company: current_user.company)
    else super end
  end
end

Formtastic::Inputs::Base::Collections.prepend(MyGlobalModule)
```

Uelb:
I think the class and configuration idea is really nice ! Do you have time to implement it or do you want me to do it ?

mikz:
@Uelb unfortunately I really don't have time to implement it. We have a 4.0 release pending for ages and I struggle to find time for that.

Uelb:
Ok, I'll do my best to submit something that won't take too much time to review

mikz:
@Uelb could you try to consider other `collection_for_` methods like for enum  https://github.com/justinfrench/formtastic/pull/1227 ? 

Also I expect some people could have already modified this behaviour, so we should not remove the `collection_for_association` method, but rather delegate to the new object and possibly print a deprecation warning like was done for the action/input finders (https://github.com/justinfrench/formtastic/pull/1063).

justinfrench:
Closing for now due to inactivity.

PR #1239, opened 2017-02-02
3.1 stable - Lambda support in collections

ebar0n:
Changes needed to be able to use lambda type collections


mikz:
Hi! I'm quite surprise this haven't worked before. Could you support it by writing a test? 

Returning the collection early means the label/id transformations won't be applied. Not sure if that is intentional or not.

Also does this exist on master that it targets 3.1 ?

justinfrench:
Closing due to inactivity and no test case. Happy to re-open or revisit with a test case (or even a comment that shows some sample code that would help someone else write a test case)!

Issue #1240, opened 2017-02-11
default_columns_for_object returns duplicates


Fivell:
refs https://github.com/activeadmin/activeadmin/issues/3828

Example 
```ruby
class Consent < ActiveRecord::Base
  belongs_to :disclosed_from_org, :class_name => "Organization", :foreign_key => "disclosed_from_orid"
  belongs_to :disclosed_to_org, :class_name => "Organization", :foreign_key => "disclosed_to_orid"
end
```

When trying to build form for such configuration  it displays duplicated inputs for each association.

After debugging I detected problematic method
```ruby
/formtastic-3.1.4/lib/formtastic/helpers/inputs_helper.rb
```

method ```default_columns_for_object```

```
 def default_columns_for_object
        cols  = association_columns(:belongs_to)
        cols += content_columns
        cols -= Formtastic::FormBuilder.skipped_columns
        cols.compact
  end
```

~~for  this case  ```disclosed_from_orid``` and  ```disclosed_to_orid``` exists in both~~

```
 association_columns(:belongs_to)
``` 
and 
```
content_columns
```

I think formtastic should skip element from content_columns  if already exits in ```association_columns(:belongs_to)```

UPD ```content_columns```  contains  ```disclosed_from_orid``` and  ```disclosed_to_orid```
and  ``` association_columns(:belongs_to)```  contains ```disclosed_from_org``` and ```disclosed_to_org``` 
that is the reason of duplicates as well

Fivell:
btw @justinfrench  maybe formtastic has possibility to override ```default_columns_for_object``` ?

Fivell:
cc @mikz

mikz:
I see. I suppose the column names are all symbols/strings ? Would it make sense to use set instead or call `.uniq` ? It is hard to imagine a scenario where you'd like to have duplicates.

Would be good to have some example schema of the model. The original issue says it shows two drop downs and two text fields which would not indicate a symbol/string array.

Fivell:
@mikz ,
> The original issue says it shows two drop downs and two text fields which would not indicate a symbol/string array.

 it shows 
dropdown for ```disclosed_to_org```   because is returned by ```association_columns(:belongs_to)```
dropdown for ```disclosed_from_org``` because is returned by ```association_columns(:belongs_to)``` 
input for  ```disclosed_from_orid```  because is returned by ```content_columns```
and input for ```disclosed_to_orid```  because is returned by ```content_columns```


[...]

PR #1241, opened 2017-02-13
Fixed default_columns_for_object when object has non-standard foreign keys

Fivell:
refs #1240

### DIFF
--- a/lib/formtastic/helpers/inputs_helper.rb
+++ b/lib/formtastic/helpers/inputs_helper.rb
@@ -350,12 +350,21 @@ module Formtastic
         end
       end
 
+      # Collects all foreign key columns
+      def foreign_key_columns # @private
+        if @object.present? && @object.class.respond_to?(:reflect_on_all_associations)
+          @object.class.reflect_on_all_associations(:belongs_to).map(&:foreign_key)
+        else
+          []
+        end
+      end
+
       # Collects content columns (non-relation columns) for the current form object class.
       def content_columns # @private
         # TODO: NameError is raised by Inflector.constantize. Consider checking if it exists instead.
         begin klass = model_name.constantize; rescue NameError; return [] end
         return [] unless klass.respond_to?(:content_columns)
-        klass.content_columns.collect { |c| c.name.to_sym }.compact
+        klass.content_columns.collect { |c| c.name.to_sym }.compact - foreign_key_columns
       end
 
       # Deals with :for option when it's supplied to inputs methods. Additional
--- a/spec/helpers/inputs_helper_spec.rb
+++ b/spec/helpers/inputs_helper_spec.rb
@@ -398,7 +398,22 @@ RSpec.describe 'Formtastic::FormBuilder#inputs' do
       it 'should not render timestamps inputs by default' do
         expect(output_buffer).not_to have_tag('form > fieldset.inputs > ol > li.datetime')
       end
-    
+
+      context "with non-standard foregin keys" do
+        before do
+          @output_buffer = ''
+        end
+
+        it 'should respect foreign key while rendering select' do
+          concat(semantic_form_for(LegacyPost.new, {:url => '/'}) do |builder|
+            concat(builder.inputs)
+          end)
+          expect(output_buffer).to have_tag('form > fieldset.inputs > ol > li.select select#legacy_post_post_author', :count => 1)
+          expect(output_buffer).not_to have_tag('input#legacy_post_post_author')
+        end
+        
+      end
+
       context "with a polymorphic association" do
         
         before do 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,5 @@
 ## master
-
+  * Fixed default_columns_for_object when object has non-standard foreign keys (#1241)
   * Fixed missing constants in production (#911)
   * Removed support for Rails 3 and 4.0 (#1108)
   * Removed deprecated input/action finder methods (#1139)
--- a/spec/schema.rb
+++ b/spec/schema.rb
@@ -13,4 +13,9 @@ ActiveRecord::Schema.define(version: 20170208011723) do
     t.datetime "updated_at"
   end
 
+  #non-rails foreign id convention
+  create_table "legacy_posts", force: :cascade do |t|
+    t.integer "post_author",  limit: 4,     null: false
+  end
+
 end
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -134,6 +134,10 @@ module FormtasticSpecHelper
     end
   end
 
+  class ::LegacyPost < ActiveRecord::Base
+     belongs_to :author, foreign_key: :post_author
+  end
+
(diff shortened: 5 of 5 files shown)


Fivell:
@mikz , failing test added

Fivell:
I will add next commit to this PR to fix this

Fivell:
@mikz , please review

spec/helpers/inputs_helper_spec.rb
```diff
@@ -398,7 +398,34 @@
       it 'should not render timestamps inputs by default' do
         expect(output_buffer).not_to have_tag('form > fieldset.inputs > ol > li.datetime')
       end
-    
+
+      context "with non-standard foregin keys" do
+        before do
+          #expect(::Post).to receive(:content_columns).and_call_original
```
mikz: This comment should not be necessary.

Fivell: @mikz , sure

Fivell:
@mikz , done, anything else ?

lib/formtastic/helpers/inputs_helper.rb
```diff
@@ -350,12 +350,21 @@ def association_columns(*by_associations) # @private
         end
       end
 
+      # Collects non-standard foreign key columns
+      def foreign_key_columns
+        if @object.present? && @object.class.respond_to?(:reflections)
+          @object.class.reflections.map {|_,v| v.options[:foreign_key] }.compact
```
mikz: Shouldn't this use just `#foreign_key` method? Because I guess we don't want to show fields for associations with no options.

Also imo it should check just `ActiveRecord::Reflection::BelongsToReflection` as the other foreign keys will not exist on this table. That should be easily checked with `Reflection#belongs_to?`

Fivell: @mikz , 

> Shouldn't this use just #foreign_key method? 

for doing this I should change all the doubles to allow foreign_key method, do you want this? 

>> Because I guess we don't want to show fields for associations with no options. 

in this case options will be empty hash and ```compact``` will ignore such columns . Also  we do care only for non-standard foreign keys which are described using ```:foreing_key```  option. 

>>Also imo it should check just ActiveRecord::Reflection::BelongsToReflection as the other foreign keys will not exist on this table. That should be easily checked with Reflection#belongs_to?

is this is ok for you ?

```ruby
 def foreign_key_columns
     if @object.present? && @object.class.respond_to?(:reflections)
         @object.class.reflections.select{|_, v| v.macro == :belongs_to }.map {|_,v|  v.options[:foreign_key] }.compact
     else
        []
     end
   end
```

mikz: I think it would be less magical to just use `Reflection#foreign_key` which will return the option or it determines the name. That sounds more robust than just checking the options as it will always try to return something even if it is name generated by rails.

Looks like there is a method from Rails `reflect_on_all_associations` that accepts macro. So it can be further simplified:

```
reflections = @object.class.try(: reflect_on_all_associations, :belongs_to) || []
reflections.map { |r| r.foreign_key.to_sym }
```

Fivell: @mikz , ok but I need to allow foreign_key for all  the doubles for this

mikz: Maybe would make sense to use correct instances of `ActiveRecord::Reflection::BelongsToReflection` and not mock them at all. That would allow checking that it works with different rails versions.

Fivell: @mikz , 

> 
Maybe would make sense to use correct instances of ActiveRecord::Reflection::BelongsToReflection

Can you add more details, I didn't get your point

Fivell: > @object.class.try(: reflect_on_all_associations, :belongs_to)

As I understand I also should stub reflect_on_all_associations to each test for this. Because specs are working with non-activerecord objects

mikz: Or include the reflection methods to those models. They are in `ActiveRecord::Reflection::ClassMethods`.

What I meant was create instance like `ActiveRecord::Reflection::BelongsToReflection.new` instead of creating mocks. That way we can really guarantee rails compatibility.

Fivell: @mikz , there is no activerecord dependency in tests now, not sure it is good idea to add it for such a small improvement

mikz: It is quite a big improvement because it ensures this will work with different rails versions. What if rails changes format of that object or that method? Because it is a mock no one can know because it does not test the rails interface. 

The tests depend on rails. They include methods from ActionView, ActiveModel and other parts. Using object from ActiveRecord and include few methods is not going to include more dependencies.

mikz: I see what you mean. Formtastic actually does not depend on ActiveRecord. But then it is kind of weak dependency as those methods like `content_columns` are from ActiveRecord.

I'd say it is fine to add ActiveRecord as a test dependency, because we actually want to test against ActiveRecord. I'll start a branch soon exploring that and replacing few mocks with ActiveRecord objects.

Fivell: @mikz, 

>> What if rails changes format of that object or that method? Because it is a mock no one can know because it does not test the rails interface.


generally I agree with you about test changes,  however  formtastic tests already over-mocked and over-stubbed and needs more efforts and time to add test for each edge case than it deserves.

>>  I'll start a branch soon exploring that and replacing few mocks with ActiveRecord objects.

feel free to ping me when it will be more clear how to arrange tests after this, so I could finish it if I have free time.

mikz: Yes formtastic's test suite is totally over-mocked and over-stubbed. I'll work over time to clean it up. 

https://github.com/justinfrench/formtastic/pull/1242 is first spark of integrating ActiveRecord.

I think you could just create some new models and add them to the schema to use real ActiveRecord objects.

Fivell:
blocked by #1241 until will be merged in

mikz:
@Fivell merged it.

Fivell:
@mikz ,  changed and  rebased

lib/formtastic/helpers/inputs_helper.rb
```diff
@@ -350,12 +350,21 @@ def association_columns(*by_associations) # @private
         end
       end
 
+      # Collects all foreign key columns
+      def foreign_key_columns
```
mikz: I guess this should be marked as ` # @private` to not show in the yard documentation

Fivell: @mikz , done

mikz:
@Fivell thank you very much! This is good change and good improvement in testing against real ActiveRecord!

Will merge it right after build is green.

Fivell:
@mikz,  thanks for help in fixing it 🚀

PR #1242, opened 2017-02-14
[tests] integrate activerecord

mikz:
* using sqlite in memory table
* real ActiveRecord models

This aims to test more Rails integration over mocking and stubbing ActiveRecord. 
Should be improving compatibility testing between different versions.


Fivell:
@mikz ,  you made changes for ```Author``` and ```::Namespaced::Post```,
what about ```::Post``` ?

mikz:
Replacing the existing instance of `@new_post` is too costly as the stubs sometimes don't make sense. It will be much longer effort to replace all uses of those stubs. But new tests can freely use new models or existing ones.

Fivell:
@mikz , ok I will add sort of ```LegacyPost``` with non standard foreign key in related PR

PR #1256, opened 2017-10-22
refactor(I18n): Improve support for i18n scope name

dhyegocalota:


mikz:
@dhyegofernando thank you for the contribution. I think this will need some explanation and a test case showing what it actually does.

dhyegofernando:
@mikz thank you

Alright, consider these models:

```ruby
class Plan < ActiveRecord::Base
  has_many :kinship_rules
  accepts_nested_attributes_for :kinship_rules
end

class KinshipRule < ActiveRecord::Base
  enum associated_type: [:personal, :company]
end
```

...this translation file:

```yaml
pt-BR:
  activerecord:
    attributes:
      kinship_rule:
        associated_type: Tipo de Associado
        associated_types:
          personal: Pessoa Física
          company: Pessoa Jurídica
```

and this admin form:

```ruby 
semantic_form_for @plan do |f|
  f.inputs do
    f.input :name
  end

  f.inputs for: :categories do |category, i|
    category.input :associated_type
  end

  f.actions
end
```


**Expected behaviour => with translated enum values**

```html
<ol>
  <li class="select input optional" id="plan_kinship_rules_attributes_0_associated_type_input">
    <label for="plan_kinship_rules_attributes_0_associated_type" class="label">Tipo de Dependente</label>
    <select name="plan[kinship_rules_attributes][0][associated_type]" id="plan_kinship_rules_attributes_0_associated_type">
      <option value=""></option>
      <option value="holder">Titular</option>
      <option value="dependent">Dependente</option>
      <option value="aggregated">Agregado</option>
    </select>
  </li>
</ol>
```


**Given behaviour => untranslated enum values**

```html
<ol>
  <li class="select input optional" id="plan_kinship_rules_attributes_0_associated_type_input">
    <label for="plan_kinship_rules_attributes_0_associated_type" class="label">Tipo de Dependente</label>
    <select name="plan[kinship_rules_attributes][0][associated_type]" id="plan_kinship_rules_attributes_0_associated_type">
      <option value=""></option>
      <option value="holder">Holder</option>
      <option value="dependent">Dependent</option>
      <option value="aggregated">Aggregated</option>
    </select>
  </li>
</ol>
```

**Why**

[These line](https://github.com/justinfrench/formtastic/blob/master/lib/formtastic/inputs/base/collections.rb#L117) is using the wrong method `object_name` because in nested forms like that, it returns `plan[kinship_rules_attributes][0]` instead of `kinship_rule`.

mikz:
Ok, this makes sense to fix. But it definitely needs a better test case, so it does not break in the future. 
What is there is just preventing it from crashing, not really testing how it does i18n.

Also, how this works with polymorphic collections?

justinfrench:
Closing for now due to inactivity and missing test case.

PR #1258, opened 2017-11-06
Allow symbols for numericality options

nikolai-b:
Rails allows symbols for these values but formtastic doesn't https://github.com/rails/rails/blob/v4.2.9/activemodel/lib/active_model/validations/numericality.rb#L55

### DIFF
--- a/lib/formtastic/inputs/base/validations.rb
+++ b/lib/formtastic/inputs/base/validations.rb
@@ -207,6 +207,8 @@ module Formtastic
 
         private
 
+        # Loosely based on
+        # https://github.com/rails/rails/blob/5-2-stable/activemodel/lib/active_model/validations/numericality.rb#L54-L59
         def option_value(option, object)
           case option
           when Symbol
--- a/lib/formtastic/inputs/base/validations.rb
+++ b/lib/formtastic/inputs/base/validations.rb
@@ -223,4 +223,3 @@ module Formtastic
     end
   end
 end
-
--- a/lib/formtastic/inputs/base/validations.rb
+++ b/lib/formtastic/inputs/base/validations.rb
@@ -73,13 +73,11 @@ module Formtastic
             raise IndeterminableMinimumAttributeError if validation.options[:greater_than] && column? && [:float, :decimal].include?(column.type)
 
             if validation.options[:greater_than_or_equal_to]
-              return (validation.options[:greater_than_or_equal_to].call(object)) if validation.options[:greater_than_or_equal_to].kind_of?(Proc)
-              return (validation.options[:greater_than_or_equal_to])
+              return option_value(validation.options[:greater_than_or_equal_to], object)
             end
 
             if validation.options[:greater_than]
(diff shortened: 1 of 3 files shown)


nikolai-b:
Any thing I can do to move this forward?

lib/formtastic/inputs/base/validations.rb
```diff
@@ -208,6 +204,19 @@ def readonly_attribute?
         def readonly_from_options?
           options[:input_html] && options[:input_html][:readonly]
         end
+
+        private
+
+        def option_value(option, object)
```
mikz: Would be good to have a comment to point to Rails source code where this is implemented.

spec/inputs/base/validations_spec.rb
```diff
@@ -338,5 +338,145 @@ class MyInput
...
   end
+
+  describe '#validation_min' do
+    before :example do
+      allow(instance).to receive(:validations?).and_return(:true)
+      allow(instance).to receive(:validations).and_return([validator])
+    end
+
+    context 'with a greater_than numericality validator' do
+      let(:validator) { double(options: { greater_than: option_value }, kind: :numericality) }
```
mikz: Would it be too hard to use real instance of `ActiveModel::Validations::NumericalityValidator` ?

spec/inputs/base/validations_spec.rb
```diff
@@ -338,5 +338,145 @@ class MyInput
       end
     end
   end
+
+  describe '#validation_min' do
+    before :example do
+      allow(instance).to receive(:validations?).and_return(:true)
+      allow(instance).to receive(:validations).and_return([validator])
```
mikz: This is stubbing the object under test. It should create a model that has validations and use that instead.

mikz:
@nikolai-b sorry for late review. I think it is good change that matches Rails implementation and improves the test coverage, but it needs few final touches.

* do not stub the object under test, rather create a model with validations
* use real validator objects from rails - so we test the Rails integration
* just add few comments to with links to the rails source code to verify the implementation is the same

nikolai-b:
@mikz updated thanks!

lib/formtastic/inputs/base/validations.rb
```diff
@@ -208,8 +204,22 @@ def readonly_attribute?
...
           options[:input_html] && options[:input_html][:readonly]
         end
+
+        private
+
+        # Loosely based on
+        # https://github.com/rails/rails/blob/5-2-stable/activemodel/lib/active_model/validations/numericality.rb#L54-L59
+        def option_value(option, object)
+          case option
+          when Symbol
```
mikz: I think both can be unified to:

```ruby
option.to_proc.call(option)
```

mikz: But this might behave different than the rails validator on private methods. So better to keep it like this.

nikolai-b:
@mikz done, thanks!

mikz:
Thank you for the contribution! Good job 👍

PR #1263, opened 2018-02-02
And ruby 2.5, update others, in travis config

lostapathy:


lostapathy:
I also removed ruby 2.0.0 from the build matrix - it has been failing on rails 4.2 for a while now and has been unsupported for 2 years.

Is it time to update the gemspec to only support ruby 2.2 going forward, or what will be the cutoff for that?

mikz:
@lostapathy I think it would be good to set minimum required version in the gemspec when dropping 2.0. 

And I think it makes sense to drop unsupported rubies when they stop working or cause too much maintenance burden, which is not the case for 2.1 yet, right? Also I think formtastic should be supporting the same rruby versions as supported versions of rails.

lostapathy:
@mikz Ruby 2.1 support ended almost a year ago, and 2.2 is slated to end next month, see https://www.ruby-lang.org/en/news/2017/04/01/support-of-ruby-2-1-has-ended/

I really like restricting gemspecs to match what we're willing to keep working on Travis, and I see little reason to keep support for unsupported rubies (or other dependencies) except in pretty extreme circumstances.  Gem developers should spend their precious time looking forward, not supporting software it's own maintainer has walked away from.

I'll update this PR to remove ruby-2.1 and update the gemspec to match.  I'll save it for another PR, but perhaps it's also time to restrict the other dependencies (mainly rails stuff) to only supported versions of rails as well?

mikz:
@lostapathy if ruby 2.1 works with supported rails versions then formtastic should support it too. And formtastic 4.x is going to support  Rails >= 4.1, so we definitely are not dropping Ruby 2.1 anytime soon. It would make sense only if it breaks and it would take too much time to fix.

Ruby >= 2.1 sounds good to me. And >= 2.2 when we drop Rails 4.2 support. But could be done faster if it breaks sooner. 

We can reconsider when supporting old versions actually causes some maintenance, but right now it really doesn't so I'd try to not break things if possible.

lostapathy:
Thanks @mikz - I have updated this as suggested.

mikz:
@lostapathy thank you very much!

PR #1265, opened 2018-05-19
Add JSON Input for models with jsonb fields.

cuzzo:
With ActiveRecord and a Postgres database, it's possible to have fields on a record that are actual JSON objects, rather than strings.

Currently, when Formtastic tries to render these fields in the admin interface, it renders them in the standard Ruby serialized hash format:

```ruby
{ "key1" => "value1" } { "key2" => "value2" }
```

This `JsonInput` allows the value of the field to be displayed as json:

```json
[{ "key1": "value1" }, { "key2": "value2" }]
```

This is helpful so that the value of the field can be edited directly, without the user/administrator 
first having to manually converting it from the Ruby serialized hash format to JSON.

### DIFF
--- a/lib/formtastic/inputs/json_input.rb
+++ b/lib/formtastic/inputs/json_input.rb
@@ -1,0 +1,37 @@
+module Formtastic
+  module Inputs
+
+    # Outputs a simple `<label>` with a `<input type="text">` wrapped in the standard
+    # `<li>` wrapper. This is the default input choice for database columns of the `:jsonb` type.
+    # You can force any input to be a string input with `:as => :json`.
+    #
+    # @example Full form context and output
+    #
+    #   <%= semantic_form_for(@user) do |f| %>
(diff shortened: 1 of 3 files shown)


lib/formtastic/inputs/json_input.rb
```diff
@@ -0,0 +1,37 @@
...
+    #       </ol>
+    #     </fieldset>
+    #   </form>
+    #
+    # @see Formtastic::Helpers::InputsHelper#input InputsHelper#input for full documentation of all possible options.
+
+    class JsonInput < StringInput
+      def to_html
+        val = self.object.send(method)
+        self.object.send("#{method}=", JSON.generate(val))
```
mikz: I don't think this is OK. A view can't just set attributes on the model.
Also, what if I want to use `JSON.pretty_generate` ?  Or different library?

justinfrench:
Closing for now due to inactivity.

PR #1269, opened 2018-09-23
Make tests/travis work with current rails versions

lostapathy:


lostapathy:
This is done and, I believe, ready to merge.  This ended up taking a little more work than expected.

* Tests were updated to work with rails 5.1 and 5.2
* Only test against rails-edge on ruby 2.4 and newer, as edge does not work with older rubies
* Remove Travis config for ruby 2.1.  

We have an indirect dependency on byebug. byebug-9 does not support ruby-2.5, and byebug-10 does not support ruby-2.1.  I'm not aware of a way to lock the byebug dep to <10 only for ruby-2.1.  Since ruby-2.5 and rails-5.2 seems more important than supporting ruby-2.1 (which quit receiving critical security updates over 18 months ago), I opted to remove the ruby-2.1 builds.  If someone has a better solution I'm open to suggestions.

mikz:
Support for Ruby 2.1 ended in April 2017. I'm good with removing the build.
https://www.ruby-lang.org/en/news/2017/04/01/support-of-ruby-2-1-has-ended/

Thank you very much for taking care!

lostapathy:
Thanks @mikz - should I update the gemspec to reflect what we're able to do CI builds on at this point, or would you rather leave that alone?  There are a couple other things I'd like to update in there and would just as soon get it all in one PR for you.

mikz:
@lostapathy I think it is not necessary to limit the version in the gemspec. That we don't test it does not mean it does not work, so I'd rather keep it unbounded.

PR #1272, opened 2018-10-04
Add a selected option to datetime_select

dkniffin:
### Why?
- Previously, as far as I could tell, there was no way to set the value of a `datetime_select`, `date_select`, or `time_select` field. 

### What changed?
- In Rails, there is a `selected` option (documented [here](https://api.rubyonrails.org/classes/ActionView/Helpers/DateHelper.html#method-i-date_select)). I decided to follow that pattern and add a `selected` option to Formtastic that does the same thing.

### DIFF
--- a/lib/formtastic/inputs/base/timeish.rb
+++ b/lib/formtastic/inputs/base/timeish.rb
@@ -85,6 +85,9 @@ module Formtastic
       #   <%= f.input :publish_at, :as => :time_select, :include_blank => true %>
       #   <%= f.input :publish_at, :as => :time_select, :include_blank => false %>
       #
+      # @example Provide a value for the field via selected
+      #   <%= f.input :publish_at, :as => :datetime_select, :selected => DateTime.new(2018, 10, 4, 12, 00)
+      #
       # @todo Document i18n
       # @todo Check what other Rails options are supported (`start_year`, `end_year`, `use_month_numbers`, `use_short_month`, `add_month_numbers`, `prompt`), write tests for them, and otherwise support them
       # @todo Could we take the rendering from Rails' helpers and inject better HTML in and around it rather than re-inventing the whee?
@@ -157,7 +160,7 @@ module Formtastic
         end
         
         def value
-          object.send(method) if object && object.respond_to?(method)
+          input_options[:selected] || object.send(method) if object && object.respond_to?(method)
         end
         
         def fragment_input_html(fragment)
--- a/lib/formtastic/inputs/base/timeish.rb
+++ b/lib/formtastic/inputs/base/timeish.rb
@@ -160,7 +160,8 @@ module Formtastic
         end
         
         def value
-          input_options[:selected] || object.send(method) if object && object.respond_to?(method)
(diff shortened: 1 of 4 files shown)


dkniffin:
Hmm. Not sure why the build failed, and I can't tell if it's failing in master or not.

mikz:
Triggered a rebuild of master: https://travis-ci.org/justinfrench/formtastic/builds/433035374
And of your branch: https://travis-ci.org/justinfrench/formtastic/builds/437237863

dkniffin:
Thank you. It looks like it passed this time. Let me know if there's anything I need to change on this.

dkniffin:
I can definitely make that change, but that's not quite accurate, is it? `input_html` will set the values on the actual HTML element, but in this case, there isn't a single HTML element to set `value` on.

mikz:
@dkniffin hmm, because it is 3 elements and each gets only part of the value. I'm really not sure about this now. I don't like these 3 input types having a different way of setting the value than rest of the inputs. But you are right, it is probably better to keep it this way.

lib/formtastic/inputs/base/timeish.rb
```diff
@@ -157,7 +160,7 @@ def fragment_label_html(fragment)
         end
         
         def value
-          object.send(method) if object && object.respond_to?(method)
+          input_options[:selected] || object.send(method) if object && object.respond_to?(method)
```
mikz: The whole line is not going to be evaluated when the object does not respond to the method. Is that expected?

dkniffin: yeah, I feel like that's right. If the model doesn't have the attribute, we shouldn't be able to set the value of it.

mikz: But this is about getting the value. If you pass the `:selected` option, it gets ignored. I don't see why it should be ignored as it is totally independent of reading the value from the model.

I believe it really should be `input_options.fetch(:selected) { object.try(method) }`.  This would be the closest to checking the `:value` in other parts of the code linked in 
https://github.com/justinfrench/formtastic/pull/1272#pullrequestreview-162524612.

dkniffin: Yeah, that's true. There could be a use-case where someone wants some custom datetime field, and needs to set the value, regardless of whether the model has an attribute or not. I'll change it.

dkniffin:
@mikz Alright, made that change. Let me know if there's anything else

mikz:
@dkniffin thank you very much for the contribution!

dkniffin:
Thank you for being quick to respond with feedback and merging quickly! :)

dkniffin:
@mikz Any chance I can get a version bump to start using this without pointing to master?

mikz:
@dkniffin, unfortunately, that is unlikely. I'd like to release 4.0 (https://github.com/justinfrench/formtastic/issues/1197), but really haven't got the time to do so.

dkniffin:
ok, good to know. Thanks!

PR #1273, opened 2018-10-07
Removes call to config option through explicit classname

chrp:
[ActiveAdmin relies on a subclassing FormBuilder](https://github.com/activeadmin/activeadmin/blob/master/lib/active_admin/form_builder.rb). For remotely related reasons I also need to set the skipped_columns option only for that subclass but the explicit call to ``Formtastic::FormBuilder.skipped_columns`` [here](https://github.com/justinfrench/formtastic/blob/ebd9b6627e1ee867347094cb7488292ce69d01a6/lib/formtastic/helpers/inputs_helper.rb#L313) prevents me from doing that. 

General question: Should I be able to copy the config when subclassing from FormBuilder to be able to alter it's behaviour? Or should there be only one global config. 

For my very specific problem: Removing the explicit reference to the class actually works quite fine (except that a future dev might have trouble finding out where ``skipped_columns`` came from)

### DIFF
--- a/spec/builder/custom_builder_spec.rb
+++ b/spec/builder/custom_builder_spec.rb
@@ -7,7 +7,7 @@ RSpec.describe 'Formtastic::Helpers::FormHelper.builder' do
 
   class MyCustomFormBuilder < Formtastic::FormBuilder
   end
-  
+
   # TODO should be a separate spec for custom inputs
   class Formtastic::Inputs::AwesomeInput
     include Formtastic::Inputs::Base
@@ -62,7 +62,7 @@ RSpec.describe 'Formtastic::Helpers::FormHelper.builder' do
           expect(builder.class).to be(MyCustomFormBuilder)
         end
       end
-      
+
       # TODO should be a separate spec for custom inputs
       it "should allow me to call my custom input" do
         semantic_form_for(@new_post) do |builder|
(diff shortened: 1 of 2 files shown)


mikz:
This looks like a very reasonable change. The change that introduced it (https://github.com/justinfrench/formtastic/commit/d3814b4e5da665a7d5b9da5035b01b970ae9f0ed) did not count that the inputs are actually included to the form builder anyway.

Could you cover it with a test?

chrp:
Happy to cover it with a test of cause. Does it make sense to subclass FormBuilder and test that I can change the setting for that subclass? Or what should the test cover in your eyes?

mikz:
The test could cover that a subclass can have different skipped columns than the parent class (FormBuilder). So yeah, creating a subclass and setting the `skipped_columns` property, then verifying it does not render those columns.

spec/builder/custom_builder_spec.rb
```diff
@@ -73,10 +73,21 @@ def to_html
       # See: https://github.com/justinfrench/formtastic/issues/657
       it "should not conflict with navigasmic" do
         allow_any_instance_of(self.class).to receive(:builder).and_return('navigasmic')
-        
+
         expect { semantic_form_for(@new_post) { |f| } }.not_to raise_error
       end
 
+      it "should use the custom builder's skipped_columns config for inputs" do
+        MyCustomFormBuilder.skipped_columns = [:title, :created_at]
```
mikz: This can affect other tests because it is done on the class level. It would be better to create an anonymous class just in this test case.

mikz:
Thank you!

Issue #1115, opened 2014-12-08
Unable to find input class InetInput


rodrigopinto:
Hi guys. I am using activeadmin on my models and I came across with a error caused by formtastic that is used behind activeadmin.

```
Formtastic::UnknownInputError at /admin/users/1/edit
Unable to find input class InetInput
```

When using a devise model, desive adds two columns as `inet`([network address type](http://www.postgresql.org/docs/9.1/static/datatype-net-types.html)), but there isn't a InetInput class declared on formtastic, so [input_helper.rb](https://github.com/justinfrench/formtastic/blob/master/lib/formtastic/helpers/input_helper.rb#L330) can not resolve the lookup for the correct class to instantiate.

mitio:
I think that the inet fields should be treated as string by default. They're used by default for trackable Devise models on PostgreSQL and it's a bit annoying to have to redefine the whole form for the resource because of these inet fields. If the form is not redefined, an exception will be raised.

timoschilling:
@justinfrench what do you thing about a `DefaultInput` or `FallbackInput`, which is used for all unknown inputs?

justinfrench:
@timoschilling @mitio We already fallback to `StringInput` in many cases. I think the problem here is that we don't have a fallback to `StringInput` when we're mapping database column types to inputs. I've created a new issue #1146 for this, would love help patching :)

PR #1276, opened 2019-02-13
Support Inet database field types

danielricecodes:
Seems people want this.  See #1146 and #1115.  Also, this [ActiveAdmin issue](https://github.com/activeadmin/activeadmin/issues/4178).

### DIFF
--- a/formtastic.gemspec
+++ b/formtastic.gemspec
@@ -40,5 +40,5 @@ Gem::Specification.new do |s|
   s.add_development_dependency(%q<rake>)
   s.add_development_dependency(%q<activemodel>, [">= 3.2.13"])
   s.add_development_dependency(%q<activerecord>, [">= 3.2.13"])
-  s.add_development_dependency(%q<sqlite3>)
+  s.add_development_dependency(%q<sqlite3>, ["~> 1.3.6"])
 end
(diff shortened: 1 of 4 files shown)


mikz:
I believe this is not the correct approach to solve this. There is no reason for this class, as it has no implementation.

The problem is that some code defaults to `column.type`, which is not correct in most cases.

https://github.com/justinfrench/formtastic/blob/2f8b987ee4880c4aa06e6ec799eafe48fa46e316/lib/formtastic/helpers/input_helper.rb#L291

We would have to define some mapping, default or something. But defining a class for every column type we want to alias is not the way to go.

danielricecodes:
Are you suggesting to add an inet switch like this?

```ruby
when :inet
  return :string
```

mikz:
Yes. And appropriate test that verifies what it does.

danielricecodes:
@mikz - simple enough.  Sorry about that empty commit, let me know if you want me to remove https://github.com/justinfrench/formtastic/pull/1276/commits/7a3649e04728ac6864cebd2b373f01a38d057bd6

danielricecodes:
@mikz - I am not sure whats going on with Travis but it appears the build is broken.  My change didn't cause it.  I can run RSpec and get 100% passing on my local system.  Ruby MRI 2.6.1 w/ RVM.  

screenshot from 2019-02-15 14-56-06

Issue #1278, opened 2019-05-28
semantic_errors helper possible edge case XSS vector.


amlisirev:
Steps to reproduce:
- have a model with f.ex. email attribute, with validation for the address.
- enter `<script>alert('test')</script>` as an email, failing the validation.
- have the error message for the field contain the user input.
- use `f.semantic_errors *f.object.errors.keys` on the form

Is there a deliberate design choice why the semantic_errors helper uses html_safe instead of sanitising the error messages? In many cases the errors contain user input from the fields that failed validation, and as such, rogue <script> tags get executed when displaying errors on model fields with. 

The issue is relevant in cases where the model object gets attributes from sources besides the form fields (such as request params or database), that the user has possibly no control of.

mikz:
I guess you refer to https://github.com/justinfrench/formtastic/blob/f94423144eee14ccc03a64267fcafd583d3cdf85/lib/formtastic/helpers/errors_helper.rb#L54-L56 right?

It seems to me that the inner call to `template.content_tag(:li, error)` should escape the error. If it does not it is definitely not intentional.

amlisirev:
Correct, that's the implementation I was referring to. I'm not intimately familiar with the html_safe implementation, is it the case that the SafeBuffer it wraps the input in bypasses all sanitisation for the target?

mikz:
Yes and no. It will mark the buffer as html safe even when it wasn't escaped. But it should be escaped already by `template.content_tag(:li, error)`. Could you check the `error` is `.html_safe?`? 

Also, I'll be happy to take a look at a fix if you'd provide a failing test case.

gkop:
Hey folks! I just wanted to note that today the latest version of Formtastic on Rubygems is 3.1.5, and in that tree there is an [inner call to html_safe](https://github.com/justinfrench/formtastic/blob/3.1.5/lib/formtastic/helpers/errors_helper.rb#L55) that could explain this behavior. Maybe part of why we're confused is that the [inner call was removed almost 5 years ago](https://github.com/justinfrench/formtastic/commit/6da1d7bf4c147af879d81b34687ca611f1197473#diff-4e4989ba548de42d5d8bf8942a5cad02R55), and this change is reflected on master, but not reflected in 3.1.5 which was released just over 2 years ago?

mikz:
@gkop that would explain it. It might have been because of some compatibility issues with older rails versions. I guess we could try to release a 4.0 beta version with what is in master now.

[...]

PR #1279, opened 2019-06-25
took off html_safe because of xss vulnerability

ti-lun:
https://github.com/justinfrench/formtastic/issues/1278


mikz:
@ti-lun thanks for the contribution! Good find. Would you mind adding a test case? Otherwise, it would have to wait for me to find some time.

PR #1281, opened 2019-07-24
Macro method usage change for Mongoid 7.0

Mar1cX:
**Issue:** Mongoid with 7.0 patch removed file which contained `macro` method. Formtastic gem relies on this method in various places in code. Currently, projects which use Mongo 7.0 are having `undefined 'macro' method` error.
  - File path which contains `macro` method: https://github.com/mongodb/mongoid/blob/v6.4.4/lib/mongoid/relations/referenced/in.rb
  - A path where removed `relations` folder and forward files - removed: https://github.com/mongodb/mongoid/tree/v7.0.0/lib/mongoid

**Solution:** To avoid breaking current macro usage and fix the error for Mongoid 7.0 database based projects, a solution with `Proxy Object` was introduced in this pull request. Using Proxy Object allowed fixing macro method issue while keeping gem mostly intact. In the future, if other special issues will appear with the reflection object, fixing them will be easier by modifying this class.

### DIFF
--- a/lib/formtastic.rb
+++ b/lib/formtastic.rb
@@ -12,6 +12,7 @@ module Formtastic
   autoload :InputClassFinder
   autoload :ActionClassFinder
   autoload :Deprecation
+  autoload :Reflection
 
   eager_autoload do
     autoload :I18n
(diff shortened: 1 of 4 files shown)


Mar1cX:
I'll do tests the next day.

PR #1283, opened 2019-08-13
Add a second scope for i18n enum to follow the syntax in Rails guides

nicolas-brousse:
In Rails guides there is the two following paragraph about I18n and ActiveRecord:
- https://guides.rubyonrails.org/i18n.html#translations-for-active-record-models
- https://guides.rubyonrails.org/i18n.html#error-message-scopes

The guides talk about this

```yaml
en:
  activerecord:
    attributes:
      user/gender:
        female: "Female"
        male: "Male"
```

Instead of 

```yaml
en:
  activerecord:
    attributes:
      user:
        genders:
          female: "Female"
          male: "Male"
```

Since this I18n scope is present since 5 years ago, my PR add a second scope instead of changing it. That way we are able to use both, and not broke existing Rails applications.

I've to add tests. I just start this as a discussion before continuing that way.

### DIFF
--- a/lib/formtastic/inputs/base/collections.rb
+++ b/lib/formtastic/inputs/base/collections.rb
@@ -114,8 +114,11 @@ module Formtastic
 
             enum_options_hash = object.defined_enums[method_name]
             enum_options_hash.map do |name, value|
-              key = "activerecord.attributes.#{object_name}.#{method_name.pluralize}.#{name}"
-              label = ::I18n.translate(key, :default => name.humanize)
+              scopes = [
+                "activerecord.attributes.#{object_name}/#{method_name.pluralize}"
+                "activerecord.attributes.#{object_name}.#{method_name.pluralize}",
+              ]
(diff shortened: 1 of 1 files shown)


justinfrench:
Hi @nicolas-brousse, sorry for not reviewing this sooner. This looks sensible, but would need tests before it's merged. Are you in a position to add these (or at least try)?

nicolas-brousse:
Hey @justinfrench, no worries, I know that it a quite sensible change. I tried to found an other way to manage it, but didn't found. So I propose it this way.

So yes, if this change sounds interesting for you, I could try to found time to add some tests.  
I already tried it temporarily in a real project. But since it's a sensible change, I prefer to work on a test.

I'll try to found some time on next week since I'm currently sick. I'll keep you in touch.

justinfrench:
Closing due to lack of activity, but I'm also pretty confident this has already been fixed through dd4c29ae.

PR #1293, opened 2020-02-25
Drop old rails support

deivid-rodriguez:
This PR drops support for old versions of Rails that are no longer supported upstream.


justinfrench:
Looks good, merging!

deivid-rodriguez:
Thanks @justinfrench! :purple_heart:

PR #1296, opened 2020-03-09
Fix appraisal generate

deivid-rodriguez:
Currently `bundle exec appraisal generate` generates some source control changes.

This change gets the result of that command in sync with what we have under version control.


mikz:
Thank you! I think it was there for some unreleased rails 5 version.

PR #1297, opened 2020-03-10
Improve CI results against rails edge

deivid-rodriguez:
This PR brings rails-edge CI entries closer to green.

There's still two failures, but I think they might be "true positives" (as in, they signal actual problems in formtastic when running against current rails edge, and not just CI setup issues like the ones being fixed here), so I prefer to investigate them separately.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,5 +22,8 @@ script:
   - bundle exec rake spec
   - script/integration.sh
 matrix:
+  exclude:
+    - rvm: 2.4.9
+      gemfile: gemfiles/rails_edge.gemfile
   allow_failures:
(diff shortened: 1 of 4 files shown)


deivid-rodriguez:
I confirmed that the two failures don't happen against the latest Rails 6 release, so after this PR adding Rails 6 to the CI should be straightforward :tada:

deivid-rodriguez:
@mikz Pinging you since you merged the last PR? Do you think you'll be able to find some time to review this? Thank you! :purple_heart:

mikz:
Thank you!

deivid-rodriguez:
Thank YOU!!

PR #1299, opened 2020-03-16
Fix eager loading problem

daniel-rikowski:
Caused by cbbff9c7 which removed `formtastic/util.rb` but not the corresponding `autoload :Util` statement.

### DIFF
--- a/lib/formtastic.rb
+++ b/lib/formtastic.rb
@@ -18,7 +18,6 @@ module Formtastic
     autoload :FormBuilder
     autoload :Inputs
     autoload :Actions
-    autoload :Util
   end
 
   # @public


deivid-rodriguez:
This looks good to me! Sorry for the issue :sweat_smile:

I wonder how this happened (just to see is there's an issue way to get the issue covered by our specs).  Are you maybe using `Formtastic.eager_load!`?

daniel-rikowski:
No problem at all. I was running my Rails 6 app in production mode. Seems to be whatever functionality is behind the `config.eager_load` flag.

mikz:
Thank you both! 🙇

deivid-rodriguez:
Thanks! :heart_eyes:

PR #1300, opened 2020-03-20
Test against Rails 6

deivid-rodriguez:
Rails 6 does not support ruby 2.4, so that combination is excluded from TravisCI.


deivid-rodriguez:
@mikz Since you're here, this one and #1301 should be straightforward too I believe.

mikz:
🤟 very nice. Thanks!

PR #1302, opened 2020-04-13
Commit all lockfiles

deivid-rodriguez:
This ensures that any contributor cloning the repo after months without repository activity still gets a deterministic set of dependencies, and as a consequence a reliably passing CI.

The rubygems team recommends this in general, but I think it's specially interesting for project with lower activity like this one.


mikz:
This also locks all the automatic rails testing in the CI. I'd prefer to not have the `gemfiles/*` in the repo, so the CI keeps working as rails release new versions (and keeps tracking edge).

deivid-rodriguez:
Yup, that's the point, and not commiting lockfiles has the exact opposite effect: it breaks the CI. With commited lockfiles, you fix issues with your dependencies and add support for newer versions of at your own pace, while keeping always a consistent green CI status, which doesn't depend on breaking releases of external dependencies.

You can read the rubygems team official recommendation here: https://bundler.io/guides/faq.html#using-gemfiles-inside-gems.

Or the position of other package management systems, like npm, here: https://classic.yarnpkg.com/blog/2016/11/24/lockfiles-for-all/.

As I said, this become even more important if a project is highly inactive and barely has resources to keep its own CI green. For this particular project, before I started working on the CI, newest versions of Rails were not even tested, and failures against the newest versions were left in `allowed_failures` so nobody would notice breakage anyways.

mikz:
I get your point and I'm all for having a Gemfile.lock for the main project. But for the generated locks by the appraisal gem I don't agree. It would give a false sense of security, that actually everything is ok when it is not. 
The `gemfiles` folder follows the appraisal gem recommendation: https://github.com/thoughtbot/appraisal#version-control
If it would be locked in git, then it would almost never update because it is not very active. 
Rails edge wasn't really a supported target, so it was in allowed failures intentionally. 

The point here is that if rails release a security update, a patch release of 5.2 for example. If it breaks formtastic, we should know. With the lockfiles in `gemfiles` we would not know.

deivid-rodriguez:
Well, not really. We would know only if someone opens a PR, a maintainer sees the PR, a maintainer goes check whether the failing status is due to the changes in the PR or something else, and then finds out.

If Rails released a security release that breaks formtastic, I'm very positive that we're going to find out because through ab end user ticket before than through the CI status of a PR. Formtastic has had like 25 PRs in the last two years, and I created like half of them :rofl:.

Anyways, I've had this conversation many times, so I'll remove the changes in `gemfiles/`.

mikz:
Oh. I thought we have cron set up to run it at some interval.
I'll set up a daily task and see if it is mostly green.

mikz:
Enabled daily builds. I suspect we might have to set rails edge back to allowed failures as that might not be all the time stable.

deivid-rodriguez:
Ok, that sounds more reasonable. What about lockfing the dependencies for PRs (where random contributorsn want to contribute their stuff), and unlocking them for cron jobs?

It'd a matter of pointing to the stable branches of Rails and lock those, like I'm doing here, and then remove the lockfiles on cron job runs.

deivid-rodriguez:
Note that Rails edge is currently in allowed failures, since it's failing. I think we should only test edge rails on cron jobs, actually.

mikz:
But contributors don't really use those dependencies in `gemfiles` folder. They are there only for the compatibility matrix. At least I never used them unless I was working on some rails compatibility issue. Always rather triggered a CI build for all rubies and all rails versions rather than to run them locally.

deivid-rodriguez:
Not sure what you mean by "they don't use them"? What I mean is, as a random contributor to a new project, I expect that the CI status will tell me if my changes are breaking things, not if the library has issues against some version of some dependency.

deivid-rodriguez:
For example, if the Rails 5.2 job fails for a PR, I would expect the CI to be telling the contributor the her changes might be ok for the last Rails version, but that she probably needs to be more careful to stay compatible with Rails 5.2 because we still support it. I think that's useful. CI being broken because some incompatibility of the latest patch level release of Rails 5.2 that nobody has cared to fix is not.

deivid-rodriguez:
I removed the rails_edge lockfile, since it's in allowed failures and currently red anyways.

If you're ok with the current version of this PR, I'm happy to work on cron jobs running against the major versions of Rails we support _without lockfiles_.

mikz:
Right,  meant for local development, sorry.

I enabled the cron job, so we get an early alert if master is red. The contributor will also see if master is red or not. 

If we want to have "human verified" pinned rails version, we would have to create a gemfile for each of those releases. And keep `~> 5.2.0` for automatic testing of new releases.
But I don't see anyone who would from time to time bump gemfiles to latest rails versions.

That sqlite3 issue you fixed is exactly something we should get notified early and try to fix it. 
But the gemfile.lock would not fix that I think. The failure was in the integration test, not really in the lockfile. And the integration test tries to create new  a app and rails does not lock sqlite3.

Hopefully the cron tasks get us an early warning next time.

deivid-rodriguez:
Mmmm, I didn't really fix any incompatibility with Rails or sqlite, I was just fighting with trying to get the CI green. Committed lockfiles would've saved me A LOT of trouble with that, and it's one of the first things I made while working on it and investigating what was going on, actually. If you look at [the original PR](https://github.com/justinfrench/formtastic/pull/1287) where I first fixed the CI, you can see how I committed the lockfiles in there.

Anyways, if I understand your expectations, I should remove all lockfiles from this PR, except for the `Gemfile.lock` file, right?

deivid-rodriguez:
> I enabled the cron job, so we get an early alert if master is red. The contributor will also see if master is red or not.

> If we want to have "human verified" pinned rails version, we would have to create a gemfile for each of those releases. And keep ~> 5.2.0 for automatic testing of new releases.
But I don't see anyone who would from time to time bump gemfiles to latest rails versions.

These sentences are a bit confusing to me but I think your point is that nobody will update the lockfiles if we commit them. I normally use a bot to update these without my intervention (if green), but even if not, I prefer that at least the CI is stable and that someone willing to contribute gets a green CI by default. Again, more specially on inactive repos where issues are not really fixed because maintainers don't know about them, but simply because nobody has the time or will to step in and work on them.

mikz:
Yes, my expectation is to have just the main lockfile in git and leave the others out. 

Master does not break that often. Honestly, I thought the cron is on and it is green. 

If you want to have "some" good configuration then feel free to commit some of the gemfile locks, for some specific rails version, but also keep the `~>` so it tests the latest rails patch release.

If master breaks and I'm notified I'll go and take a look and try to fix it :)

deivid-rodriguez:
> If you want to have "some" good configuration then feel free to commit some of the gemfile locks, for some specific rails version, but also keep the ~> so it tests the latest rails patch release.

What you mean by "some"? Which ones? What you mean by "feel free to commit"? I don't have permissions on this repo. What you mean by "keep the ~>`? I'm not changing that in this PR. I'm super confused :rofl:.

Anyways, I updated the PR as requested.

I really really recommend reading https://classic.yarnpkg.com/blog/2016/11/24/lockfiles-for-all/ in depth before making an opinion about this. It explains really well why gitignoring lockfiles does not really help with finding issues, how it harms the contributor experience.

mikz:
@deivid-rodriguez I'm not against commiting lockfiles. I'm against not testing latest rails releases when they are out. As long as that does not break, I don't care about the lock files.
And for development and apps it is needded, I agree. But for testing a library works with the latest rails release we need something more fluid.

By feel free to commit I meant to this branch or to another PR :)

PR #1304, opened 2020-04-15
Fix dependency caching

deivid-rodriguez:
This PR fixes caching of dependencies on TravisCI, taking down 1 minute per job.

Unfortunately, it is not possible to fix this with appraisals, because:

* Appraisals enforces a shared root for all gemfiles (the `gemfiles/` folder).
* TravisCI runs `bundle clean` right before saving the cache for a given gemfile, and that messes up the cache for the other gemfiles, since the location is shared for all of them.

Fortunately, `appraisals` doesn't really do much, just loop through all gemfiles to run `Rake` tasks on them. So I implemented the most common appraisal tasks, namely, run tests for all gemfiles, and run `bundle install` for all gemfiles from scratch instead.


deivid-rodriguez:
@mikz Are you ok with dropping appraisals? I can reimplement other common tasks that you're used to if necessary, it should be fairly easy.

Some tasks I'd rather no reimplement though them because I think they are overkill. For example:

* Gemfile generation. With appraisals, in our current situation, if you want to add a dependency, you edit the `Appraisals` file in as many places as gemfiles you want to edit, and then you run a command to transmit the changes to the actual gemfiles. Without appraisals, you simply edit the gemfiles. I believes both options are similar in complexity? Also, given how simple the gemfiles are (currently they only include `rails`), editing dependencies happens very very rarely. And also note that with `appraisals` you need to keep everything in sync and it's very likely to get out of sync. As a matter of fact, I just found out that our `Appraisals` file had leftover definitions from rails 4.1 and other no longer supported Rails versions.
  
* Running tasks for a single `Gemfile` through, for example, `bundle exec appraisal rails-5.2 rake foo`. I believe it's as simple to do `BUNDLE_GEMFILE=gemfiles/rails_52/Gemfile bundle exec rake foo`. And when you are working against a specific `Gemfile`, you normally run many commands against it, so I think it's better to run `export BUNDLE_GEMFILE=gemfiles/rails_52/Gemfile` and then run commands normally, rather than having to prefix `appraisal rails-5.2` to every single command. Also, doing that matches exactly what TravisCI does, so it reduces differences between local development and CI.

Thoughs?

Rakefile
```diff
@@ -13,6 +12,24 @@ else
...
+end
+
+namespace :appraisal do
+  task :install do
+    for_all_gemfiles("install")
+  end
+end
+
+def for_all_gemfiles(*args)
+  Dir.glob("gemfiles/rails_{5.2,6.0,edge}/Gemfile").each do |gemfile|
```
mikz, line 26: Would be better to have just `gemfiles/*/Gemfile` no? So it does not need to change when we change compat matrix.

Rakefile
```diff
@@ -13,6 +12,24 @@ else
...
+    Bundler.with_original_env do
```
mikz, line 27: Rdoc says `with_original_env` is deprecated in favour of `with_unbundled_env`.

deivid-rodriguez: Mmmm, tell me where please because I should fix that since it's incorrect :grimacing:. `Bundler.with_clean_env` is the one deprecated in favor of `Bundler.with_unbundled_env`.

Rakefile
deivid-rodriguez: Yes, that's what I wrote initially, but that was running `rails_edge` first (which fails), so I figure it would be better to enforce an order where stable versions come before unstable ones.

Rakefile
mikz: Also, there is `unbundled_exec` in latest bundler.

Rakefile
mikz: I see. It does not run all tasks you mean? 

The order should not matter. If you want to evaluate all tasks before exiting with a status, then it should do something else like:
```ruby
gemfiles.map do |gemfile|
  sh(...) { |ok, status| next status }
end.all(&:success?)
```

deivid-rodriguez: Maybe I can do something like

```ruby
 [*Dir.glob("gemfiles/rails_[5-9].[0-9]/Gemfile"), "gemfiles/rails_edge/Gemfile"]
```

That wouldn't require updating until Rails 10.

Rakefile
mikz: My bad, confused those two :)

Rakefile
mikz: I don't think we should impose naming conventions on folders.

I think there are two problems. One is that it exits on the first failure.

And the second is that it runs rails edge first. 
Wouldn't calling `.sort` sort the array? Or you could give it a block, right.

Rakefile
```diff
@@ -13,6 +12,24 @@ else
   task :default => :spec
 end
 
+task :appraisal do
+  for_all_gemfiles("exec", "rake")
+end
+
+namespace :appraisal do
+  task :install do
```
mikz, line 20: It might be nice to actually generate a task for each appraisal. 
So they are visible when executed with `rake -T`. Also. this should have `desc` so it appears there with a description.

Rakefile
deivid-rodriguez: Exiting on first failure is the current behavior of the current default task, so I preserved it here. If we want to change that for some reason, I think it should be discussed and done on a separate PR.

Regarding the order, sorting works and feels better than my approach! I'll update it.

deivid-rodriguez: Updated.

Rakefile
deivid-rodriguez: This is what I explain in the second bullet point in https://github.com/justinfrench/formtastic/pull/1304#issuecomment-614248433 that I don't think it's worth it, I believe.

I'll add a `desc`! :+1:

deivid-rodriguez: Updated!

deivid-rodriguez:
Thanks for all your feedback @mikz! I marked both converstations as resolved but if you have further feedback, go ahead and unresolve them :)

Rakefile
mikz: Cool 👍

deivid-rodriguez:
Thanks for your patience and feedback @mikz! :purple_heart:

mikz:
You are making an excellent job 🥇 👍  Thank you.

PR #1309, opened 2020-05-28
Remove ruby 2.7 deprecation warning

deivid-rodriguez:
```
/path/to/formtastic-f95dddae2fe5/lib/formtastic/i18n.rb:26: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
/path/to/i18n-1.8.2/lib/i18n.rb:195: warning: The called method `translate' is defined here
```


deivid-rodriguez:
Hi @mikz and @justinfrench! Would you have some time to review this fix? Thanks!! :purple_heart:

PR #1321, opened 2020-12-23
Replace TravisCI with Github Actions

deivid-rodriguez:
TravisCI has limitations making it no longer usable for opensource projects.


deivid-rodriguez:
I created #1322, same thing but from the main repository, not from my fork, so that Github Actions trigger properly.

PR #1322, opened 2020-12-23
Replace TravisCI with Github Actions

deivid-rodriguez:
TravisCI has limitations making it no longer usable for opensource projects.


deivid-rodriguez:
I'll merge this since TravisCI will shut us down "in a few weeks".

mikz:
@deivid-rodriguez do you have any idea, why the action is not executed on some pull requests from forks? 

For example, https://github.com/formtastic/formtastic/pull/1343 does not have a run, but https://github.com/formtastic/formtastic/pull/1339 does.

deivid-rodriguez:
I have no idea @mikz, really weird... :thinking:

PR #1332, opened 2021-01-25
Drop ruby 2.4 support

deivid-rodriguez:
It reached it's End of Life in Abril 2020:
https://www.ruby-lang.org/es/news/2020/04/05/support-of-ruby-2-4-has-ended/.


deivid-rodriguez:
Ruby 2.4 is too old, I will merge this in a couple of days.

PR #1335, opened 2021-02-23
Set relevant branches to main in rails_edge appraisals Gemfile

AMHOL:


AMHOL:
NVM, was trying to run `bin/appraisal install` under ruby 3, which caused it to fail

PR #1336, opened 2021-03-12
Determine NamespacedClassFinder#finder method after rails config

s-mage:
Hi there!

You're making a great gem, thank you very much.

I've found a bug while profiling a thing in our app.

Formtastic has two modes of finding classes, one is fast and another fetches new files in development. It works in dev mode all the time because `use_const_defined?` is called when `Rails.application` is still nil.

As a result, finding input classes sometimes takes half of the time Formtastic works.

This PR fixes this by determining the finder method on the first calling, which likely happens after the app initialization.

Bonus: spec for the class tested changed `cache_classes` config value, which, I suppose, was used before `eager_load`.

Testing
-------

I don't know how to make a spec for the case, but here is how I do it manually:
1. `config.eager_load = !!ENV["EAGER_LOAD"] # config/environments/development.rb`
2. `EAGER_LOAD=true rails c`
3. For master: 
```
show-source Formtastic::NamespacedClassFinder#finder # for pry
Formtastic::NamespacedClassFinder.new([]).method(:finder).source_location # would work everywhere
```
For this PR:
```
Formtastic::NamespacedClassFinder.send(:finder_method)
```

Workaround
------------

Just in case the fix won't get to master or will be released only for v4. I've added a workaround to reach the same effect
```
# config/initializers/formtastic_namespaced_class_finder_patch.rb
if Rails.configuration.eager_load
  class Formtastic::NamespacedClassFinder
    private
      def finder(class_name)
        find_with_const_defined(class_name)
      end
  end
end
```

### DIFF
--- a/lib/formtastic/namespaced_class_finder.rb
+++ b/lib/formtastic/namespaced_class_finder.rb
@@ -32,6 +32,10 @@ module Formtastic
       defined?(Rails) && ::Rails.application && ::Rails.application.config.respond_to?(:eager_load) && ::Rails.application.config.eager_load
     end
 
+    def self.finder_method
+      @finder_method ||= use_const_defined? ? :find_with_const_defined : :find_by_trying
+    end
+
     # @param namespaces [Array<Module>]
     def initialize(namespaces)
       @namespaces = namespaces.flatten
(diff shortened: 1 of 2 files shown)


mikz (approved):
Makes sense and it looks good. The failing tests were fixed in a different PR. Thanks for this!

PR #1337, opened 2021-05-04
Stop using deprecated master branch for some gems

tomgilligan:


deivid-rodriguez:
This should be unnecessary if you use bundler >= 2.2.5 or higher, which includes https://github.com/rubygems/rubygems/pull/4224.

tomgilligan:
I was running against Ruby 2.7.2 (with an old default bundler) because Rails 5.2 appraisal needs old keyword args behaviour and didn't know about new bundler behaviour.  Running with a newer version of bundler makes this PR unnecessary.  Thank you!

PR #1338, opened 2021-05-04
Use SafeBuffer to better mimic Rails in tests

tomgilligan:
Small, mostly inconsequential change but makes the tests a bit more accurate in replicating what `output_buffer` looks like.


tomgilligan:
This work is included in #1339 , it'll be easier to understand if I just close this I think.

PR #1339, opened 2021-05-04
Avoid mutating string literals

tomgilligan:
- Makes the tests a bit more accurate in replicating what `output_buffer` looks like in a Rails app.
- Adds `# frozen_string_literal: true` to all `*.rb` files
- Replaced string literal mutation in as many places as possible

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    formtastic (4.0.0.rc1)
+    formtastic (4.0.0)
       actionpack (>= 5.2.0)
 
 GEM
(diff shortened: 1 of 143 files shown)


tomgilligan:
Needs #1340 for green build

mikz (commented):
I think this looks good.

I'm a tiny bit worried that this might affect some user code. Like custom inputs & actions. If someone inherits & extends those they would need to stop mutating strings too. 

Have you had any issues with this in activeadmin?

mikz:
@tomgilligan hey, I pushed to your PR.

Reverted the change of string concatenation instead of mutation. Rails methods like `content_tag` return mutable strings, so all code paths can rely on them. We should just keep doing the same thing in all code paths. 
It could be a massive change for user code that actually relies on these mutable strings.

PR #1342, opened 2021-06-04
Fix I18n lookup for enum values in nested select fields

### LINKED ISSUE #1151: collection_from_enum I18n unusable with has_many nested forms

Fs00:
This PR accomplishes the same as #1227, albeit with a more focused approach.
When constructing the key for looking up the localized enum value, we now use [`ActiveModel::Name.i18n_key`](https://api.rubyonrails.org/classes/ActiveModel/Name.html) instead of `object_name`, which returned something like `user[posts_attributes][0]` in case of nested forms.
I've added a test that reproduces the exact scenario that triggered the fixed bug (#1151).

Bonus: I've added a commit to update Formtastic version in Gemfile.lock (it was still 4.0.0.rc1).  

Fixes #1151
Supersedes #1227

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    formtastic (4.0.0.rc1)
+    formtastic (4.0.0)
       actionpack (>= 5.2.0)
 
 GEM
--- a/lib/formtastic/inputs/base/collections.rb
+++ b/lib/formtastic/inputs/base/collections.rb
@@ -114,7 +114,7 @@ module Formtastic
 
             enum_options_hash = object.defined_enums[method_name]
             enum_options_hash.map do |name, value|
-              key = "activerecord.attributes.#{object_name}.#{method_name.pluralize}.#{name}"
+              key = "activerecord.attributes.#{object.model_name.i18n_key}.#{method_name.pluralize}.#{name}"
               label = ::I18n.translate(key, :default => name.humanize)
               [label, name]
             end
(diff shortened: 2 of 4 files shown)


Fs00:
@mikz @justinfrench Any plans to review/merge this in the near future?

mikz (approved):
Sorry, looks like I'm not subscribed to new issues 🤦🏻‍♂️

I think this looks great! It does not break tests, it adds a new one, it is very clear in what it does.

Great fix 👍

Fs00:
Thanks!
Will there be a new release anytime soon? I think that there have been some interesting fixes since 4.0.0 that may be worth a patch release.

mikz:
@Fs00 indeed, but as we are not doing many there is always some element of fear included :) 
I'd recommend pointing bundler to the git repo in the meantime.

Fs00:
Oh well, less changes, less fear! 😄 I think that waiting will only make things more risky and difficult.

Anyway, I cannot point Bundler to the Git repo, since I use this gem in an enterprise environment where dependencies can't be pulled from the "outside world". I'll wait patiently 🤷‍♂️

PR #1343, opened 2021-06-22
Support scopes in relationships

oboxodo:
Rails has replaced the `:conditions` relationships' option with optional lambda/proc `scopes` for a long time already. It just makes sense for Formtastic to support them the same way it does for the `:conditions` option.

Given this code:

```ruby
class Post < ApplicationRecord
  belongs_to :author, -> { active.order(name: :asc) }
end

class Author < ApplicationRecord
  scope :active, -> { where(active: true) }
end
```

```slim
- semantic_form_for(Post.new) do |f|
  = f.input :author
```

Without this change the autogenerated select input with the list of authors will include ALL authors because it effectively loads the list from `Author.all`. After this change, the list will be filled with `Author.active.order(name: :asc)` because the `belongs_to` relationship includes a scope.

This seems to bring parity to the old way of applying conditions to relationships via the `:conditions` option which got deprecated back in Rails 4.

### DIFF
--- a/lib/formtastic/inputs/base/collections.rb
+++ b/lib/formtastic/inputs/base/collections.rb
@@ -78,6 +78,8 @@ module Formtastic
                     ) if reflection.options[:polymorphic] == true
             end
 
+            return reflection.klass.merge(reflection.scope) if reflection.scope
+
             conditions_from_reflection = (reflection.respond_to?(:options) && reflection.options[:conditions]) || {}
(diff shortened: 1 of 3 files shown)


mikz (approved):
I think this is great 👍

mikz:
I'm happy to merge once I figure out why there is no Github Action build for this PR.

oboxodo:
Thanks! 🤘

mikz:
Thank you! 🥇 👍

PR #1344, opened 2021-08-31
Create CI job that aggregates matrix status

mikz:
It is useful to have just one commit status with the action result. Right now there is a status for each matrix job.


mikz:
`ci` is now required check to merge a PR (not including administrators)

PR #1351, opened 2022-01-22
Fix bug where StringInquirer column gets file input.

gregschmit:
I had an issue recently where I have a `:string` database column that I cast to `StringInquirer` like this:

```ruby
attribute :configuration, :string_inquirer
```

and with this attribute definition:

```ruby
class StringInquirerType < ActiveRecord::Type::String
  def type
    return :string_inquirer
  end

  def cast(value)
    return value.inquiry
  end
end
ActiveRecord::Type.register(:string_inquirer, StringInquirerType)
ActiveModel::Type.register(:string_inquirer, StringInquirerType)
```

The form input that this resolved to was the `FileInput`, because `FileColumnDetection#is_file?` returns `true`, because string inquirers responds to `file?`. I thought about removing that, but I assume there are good reasons for those `file_methods`, so rather than messing with that, I just changed the method to return false for columns where the type ends up as `ActiveSupport::StringInquirer`. This seems sensible to me, because string inquirers are (mostly) unique in always responding to inquisitive methods, so this doesn't seem to carry any risk of the list of exceptions growing to be unmanageable.

This MR applies that change, so `StringInquirer` objects don't get a `FileInput` field, rather they get `StringInquirerInput`, which is ideal for me because then my app can define the default behavior for that kind of input.


justinfrench (commented):
Hey @gregschmit thanks for contributing this! This PR is from a long time ago (sorry), but I'm cleaning things up a bit.

I don't think this is truely a bug.

1. As far as I can tell, you can still get the desired input type with eg `as: :string`. Inferring the correct input type was never intended to be perfect, it's just trying to save you from specifying a type in most cases.
2. `file_methods` (which includes `:file?` that's causing you issues) is configurable, so you can exclude it in your app
3. I'm not sure how common it is to cast a field like this, but if it's common, we probably should consider solving this in a more generalised way, where `StringInquirer` objects default to a `StringInput`.

If you think I've missed something important and/or want to push for an enhancement, I'm happy to work with you on that (we're working on v6.0, so now is a good time to introduce breaking changes or surprises).

PR #1360, opened 2022-07-14
Formtastic version vpdate from 4.0.0 up to 4.0.1

tomoyoi:
# Trouble
A vulnerability has been detected during rack is updated due to Formtastic.

# What 
Could you update Formtastic version?


h-inouchi (approved):

tomoyoi:
According to amazon inspector, it shows there is a vulnerability in rack and it is affected by formtastic.
I am not sure what exactly rack is affected by but it would be a great help if you have a look and give me some advice. 
Screen Shot 2022-07-15 at 11 20 49 AM

PR #1371, opened 2023-09-13
Rails 7.1 support

lsylvester:
Updates to support Rails 7.1.0 which is now in beta:

* Add Gemfile for 7.0 as edge is now 7.1
* Avoid use of the ActiveSupport::Deprecation singleton (which is deprecated). Use a dedicated instance of ActiveSupport::Deprecation  instead.
* Replace ActiveSupport::SafeBuffer with ActionView::OutputBuffer as  ActiveSupport::SafeBuffer  no longer has a concat method. ActionView::OutputBuffer  is not a string, so it is required to explicitly cast it to a string in places. The content of the ActionView::OutputBuffer could not be replaced, so new instances were created instead.
* Remove silencing of deprecation for `column_for_attribute` - which was related to upgrading to Rails 5 and is no longer required.


tomascco:
@lsylvester I'm not a maintainer or anything, but I think you could also update the CI at `.github/workflows/test.yml` to run with Rails 7.0 and 7.1.

For this I think you also need to run `bundle update` at the root of the gem to update the current `Gemfile.lock` to the Rails 7.1.

davidwessman (approved):
Looked through all the changes and it looks great, did you try to run the tests on Rails 6.0 as well?

lsylvester:
> did you try to run the tests on Rails 6.0 as well?

Test the tests have run against all the Gemfiles.

>  I'm not a maintainer or anything, but I think you could also update the CI at .github/workflows/test.yml to run with Rails 7.0 and 7.1.

Good catch. I have updated the test.yml to include the Rails 7.0 and Rails 7.1 gemfiles (I also added a Rails 7.1 gemfile as it has been released since the PR was originally made). I also had to relax the version of rspec-mocks to get a fix for https://github.com/rspec/rspec-mocks/pull/1514

javierjulio:
@lsylvester I'm not a maintainer. I've asked someone who is and they've approved CI so you should be able to review and address test failures. Thank you for doing this!

lsylvester:
CI was failing due to script/integration.sh script, and I had only verified the that `rspec spec` ran.

The issue is that password is not an expected type, which now raises an error since https://github.com/rails/rails/pull/42311. I have made a change to not pass in the attribute type which should make it pass, and the generator doesn't seem to use the type so I think this should be sufficient.

lsylvester:
Latest CI run failed due to 

```
Fetching https://github.com/rack/rack.git
fatal: Needed a single revision
Git error: command `git rev-parse --verify master` in directory
/home/runner/work/formtastic/formtastic/vendor/bundle/ruby/3.0.0/cache/bundler/git/rack-a69799d848c55dca271cda2858f0ae7d1dd533b8
has failed.
Revision master does not exist in the repository
https://github.com/rack/rack.git. Maybe you misspelled it?
```

I am not sure why this was working on the ruby 2.7.2 run - but this has been fixed by using the main branch for rack.

deivid-rodriguez:
Hei! Sorry for not attending this earlier.

I think that failure you got also exists in the current master branch. The problem is that `ruby/setup-ruby` uses whatever Bundler version that comes installed by default with each Ruby in absence of a lockfile. And Bundler version installed with Ruby 3.0 didn't handle git gems fine when their default branch was not named "master".

Your fix works too, but I'd rather go with using the latest Bundler everywhere to hopefully avoid similar issues in the future, so I proposed that separately to get a green baseline here: https://github.com/formtastic/formtastic/pull/1373.

deivid-rodriguez (approved):
This looks great to me!

I would remove the latest commit since it should no longer be necessary after #1373. I don't think it will cause any harm either since I bet these gems are not going to rename their default branch again, but still. Anyways, non blocking!

deivid-rodriguez:
Thanks @lsylvester, merging this!

@mikz @justinfrench Any chance we can get a release out with Rails 7.1 support?

lsylvester:
Thanks @deivid-rodriguez. I really appreciate your help with this.

justinfrench:
Thanks @lsylvester @deivid-rodriguez 

I've shipped a new 5.0.0 gem release

PR #1379, opened 2024-02-09
adding a label_html option to customize label tag

dscataglini:
This is something I monkey patch routinely in my projects.
It allows to overwrite `:class`, `:for` attributes, as well as adding custom attributes like `data-*`.
I find this very useful for it to work with stimulus and javascript frameworks/plugins.
I added specs where I thought it made sense. Not sure if you guys require to test each input individually


justinfrench:
Sorry for the delay on this, looks great, merging on 🟢

PR #1372, opened 2023-10-12
Github Actions: Updates rubies and rails

davidwessman:
- Updates actions
- Updates ubuntu version

PR #1382, opened 2024-02-22
Add support for country_select v9. Fixes #1381

### LINKED ISSUE #1381: Country input crashes with country_select >= 9.0.0
When using with country_select >= 9.0.0 we get the following error:

```
ActionView::Template::Error
wrong number of arguments (given 4, expected 1..3)
```

This is due to the method signature of `country_select` form builder having changed via https://github.com/countries/country_select/commit/021845dafffb65bb51bb6ebc0e40c540a9e83b1a

I could work on a fix for this however I would like to discuss first, should both the old and the new version be supported, or should an error be raised for versions <= 9.0.0, effectively forcing an upgrade?

leoarnold:
The `country_select` v2 syntax was introduced in [August 2014](https://github.com/countries/country_select/blob/19072737cc2b2a12723b2dd273263b71fd5395ed/UPGRADING.md).

Now, 10 years and 7 major versions later, the project [dropped support for v1 syntax](https://github.com/countries/country_select/pull/224) starting with v9.

This change adapts `formtastic` to the new method signature, removes all claims that it supports v1 syntax, and updates the URLs to the new home under a new maintainer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -606,7 +606,7 @@ You can use your new input with `:as => :date_picker`.
 
 There are none other than Rails itself, but...
 
-* If you want to use the `:country` input, you'll need to install the [country-select plugin](https://github.com/stefanpenner/country_select) (or any other country_select plugin with the same API). Both 1.x and 2.x are supported, but they behave differently when storing data, so please see their [upgrade notes](https://github.com/stefanpenner/country_select/blob/master/UPGRADING.md) if switching from 1.x.
+* If you want to use the `:country` input, you'll need to install the [country-select plugin](https://github.com/countries/country_select) (or any other country_select plugin with the same API).
 * There are a bunch of development dependencies if you plan to contribute to Formtastic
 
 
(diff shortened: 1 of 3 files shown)


davidwessman:
Looks great! 🙂

leoarnold:
I think this is waiting on #1372. Let me know when anything needs to be amended.

leoarnold:
@justinfrench

justinfrench:
Sorry @leoarnold, thanks for the nudge. I’m finding it very difficult to keep up with maintenance on this gem, as I no longer use it or even work regularly with Ruby. 

Since there are no  co flicks and the diff is straightforward I will merge this, but I have no timeline for another release sorry.

leoarnold:
@justinfrench No problem, I kinda got that from context. I think you should state that at the top of the README as well (or even archive the project?) so passers-by come with the right expectation.

justinfrench (approved):
Thanks for your patience. LGTM!

PR #1388, opened 2025-03-18
Accessibility improvements for inline errors and semantic_errors

TimeTravelerFromNow:
This PR attempts to improve accessibility of Form errors when inline errors are being rendered, and gives `semantic_errors` links to errored inputs. The goal of the PR is to maintain existing functionality and backward compatibility while providing some enhancements to Forms that have errors

1. The `semantic_errors` method is updated to render list items, and when the error is a model attribute, automatically render itself as a link to the errored input.
2. Add ids to inline errors, and automatically update `aria-describedby` the inline error id, and `aria-invalid` set to true.

**Semantic errors**
Followed this gov uk guide to inform improvements of the `semantic_errors`
https://design-system.service.gov.uk/components/error-summary/

**Inline errors**
This guide informed my decision to add aria attributes 
https://www.smashingmagazine.com/2023/02/guide-accessible-form-validation/#the-error-message

The inline error paragraph is updated for the most common error method, `error_sentence_html`, such that the id is automatically set to `"#{method_error}"`
The `input_html_options` when `errors?` is `true` become: `aria-invalid=true`, `aria-describedby="inline_error_id"`

**Code sandbox**

Screenshot 2025-03-17 at 6 51 06 PM

### DIFF
--- a/lib/formtastic/helpers/errors_helper.rb
+++ b/lib/formtastic/helpers/errors_helper.rb
@@ -42,23 +42,31 @@ module Formtastic
       def semantic_errors(*args)
         html_options = args.extract_options!
         args = args - [:base]
-        full_errors = args.inject([]) do |array, method|
-          attribute = localized_string(method, method.to_sym, :label) || humanized_attribute_name(method)
-          errors = Array(@object.errors[method.to_sym]).to_sentence
-          errors.present? ? array << [attribute, errors].join(" ") : array ||= []
+        full_errors = {}
+        args.each do |attribute|
+          full_errors[attribute] = error_message_for_attribute(attribute)
         end
-        full_errors << @object.errors[:base]
-        full_errors.flatten!
-        full_errors.compact!
-        return nil if full_errors.blank?
+        full_errors.compact_blank!
+        base_errors = @object.errors[:base]
+        base_errors = Array(@object.errors[:base]) if base_errors.is_a?(String)
+        return nil if full_errors.blank? && base_errors.blank?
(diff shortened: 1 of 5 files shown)


TimeTravelerFromNow:
@justinfrench would you please review this PR? Thanks

justinfrench (changes requested):
Thanks for your patience and persistence with this one, and thanks for considering backwards compatibility.

If I'm understanding correctly, the one _surprise_ developers may get when upgrading is that errors associated with an attribute will now be an `li a` instead of just an `li`, is that right? Can you consider what it would take to make this a feature developer can opt-out of (or opt-in to) via either an argument or configuration option.

I know the inner workings of `semantic_errors()` was already a bit of a mess, but it's now very clearly combining together two kinds of errors (errors on the base, and errors on each attribute). If you're up for it, I think we need a refactor here. Eg  `semantic_errors()` should simply create a `<ul>` that wraps two concatenated sets of list items, with the logic for each extracted into supporting methods like `semantic_error_list_from_base()` and `semantic_error_list_from_attributes(*args)`, right? Tackling this might make it much easier to handle the complexity of opting in/out of links.

justinfrench (changes requested):
This is looking good & backwards compatible, thanks! I'm not sure if we need docs _in the README_, but certainly we need a little bit of docs _somewhere_. Please add an entry to the CHANGELOG too.

justinfrench (approved):
This is looking really good, thank you!

PR #1389, opened 2025-05-18
[WIP] Fix development dependencies

justinfrench:
The dependency management is pretty dusty and crusty. This PR seeks to improve a few things to aid development:

- Resolving a compatibility issue between various Rails and SQLite gem versions
- Upgrading to a modern Appraisal setup
- Removing support for Rails versions that are no longer supported by Rails
- Raising the min Ruby & Gems versions to match Formtastic's minimum Rails version
- Adding an appraisal for Rails 8
- Sorting out various deprecation notices


formtastic.gemspec
```diff
@@ -27,6 +27,9 @@ Gem::Specification.new do |s|
 
   s.add_dependency(%q<actionpack>, [">= 6.0.0"])
 
+  s.add_development_dependency(%q<activesupport>, [">= 6.0.0"])
+  s.add_development_dependency(%q<activerecord>, [">= 6.0.0"])
```
deivid-rodriguez: I think the reason for these development dependencies not being listed is that Rails is included in each `Gemfile` in `gemfiles/rails_<version>/Gemfile`. I normally don't use `add_development_dependency` and prefer to keep development dependencies in Gemfiles, because it's more flexible (you can change sources, or use a different Rails requirement for each Gemfile). But I'm cool with your choice!

Appraisals
```diff
@@ -0,0 +1,20 @@
...
+  gem "activerecord", "~> 8.0.0"
+  gem "sqlite3", "~> 2.4.0"
+end
+
+appraise "rails-edge" do
+  gem "actionpack", git: "https://github.com/rails/rails.git", branch: "main"
+  gem "activesupport", git: "https://github.com/rails/rails.git", branch: "main"
+  gem "activerecord", git: "https://github.com/rails/rails.git", branch: "main"
+  gem "sqlite3", "~> 2.4.0"
+end
```
deivid-rodriguez, line 17: I stopped using `Appraisals` because it felt it was not doing a lot more than setting `BUNDLE_GEMFILE` for each `Gemfile` you can to test against and add some tasks to do some looping over gemfiles. But again, you did all this work to bring things up to date which is super appreciated so I'm happy to go back to `Appraisals` if you find it better.

.gitignore
```diff
@@ -14,5 +14,6 @@ log/*
 .yardoc
 doc/
 tmp
+Gemfile.lock
```
deivid-rodriguez, line 17: I normally include lockfiles in version control, also for gems, to have a more consistent development experience. But in this case, the main Gemfiles in `gemfiles/` folder were already gitignore so my practice was not even being followed so I'm cool with gitignoring all lockfiles.

Regarding security scanner warnings, I think the issue there was including the lockfile in the packaged gem, not including the lockfile in source control. But stopping including it in source control also fixes the issue as a side effect 👍.

deivid-rodriguez (approved):
Thanks for bringing stuff up to date!

I added some comments about why some things were done the way they were done previously, but nothing blocking. This is great work regardless!

formtastic.gemspec
justinfrench: Interesting…

I’m relying pretty heavily on [this post](https://yehudakatz.com/2010/12/16/clarifying-the-roles-of-the-gemspec-and-gemfile/) as a decent source of truth:

> When developing a gem, use the [gemspec](http://gembundler.com/rubygems.html?ref=yehudakatz.com) method in your Gemfile to avoid duplication. In general, a gem's Gemfile should contain the Rubygems source and a single gemspec line. Do not check your Gemfile.lock into version control, since it enforces precision that does not exist in the gem command, which is used to install gems in practice. Even if the precision could be enforced, you wouldn't want it, since it would prevent people from using your library with versions of its dependencies that are different from the ones you used to develop the gem.

justinfrench:
> Just wondering if it's worth to keep running it just for the main version of Rails?

I was thinking the same. Leave it with me!

formtastic.gemspec
deivid-rodriguez: This is mostly Bundler's core team recommendation: https://bundler.io/guides/faq.html#using-gemfiles-inside-gems. But there's upsides and downsides of each approach, it's fine that each developer uses what works best for them.

Issue #1367, opened 2023-03-01
couldn't find file 'formtastic_ie6' with type 'text/css'


hww:
Following the REAME.md there is error in the line i--- e6.css file

/*
 *= require formtastic_ie6
 */

What I do wrong?

There is the log

```
couldn't find file 'formtastic_ie6' with type 'text/css'
Checked in these paths: 
  /home/valery/projects/veda/tc-tech-platform/tc-tech-server/app/assets/config
  /home/valery/projects/veda/tc-tech-platform/tc-tech-server/app/assets/images
  /home/valery/projects/veda/tc-tech-platform/tc-tech-server/app/assets/stylesheets
  /home/valery/.local/share/gem/ruby/3.2.0/gems/jquery-rails-4.5.1/vendor/assets/javascripts
  /home/valery/.local/share/gem/ruby/3.2.0/gems/stimulus-rails-1.2.1/app/assets/javascripts
  /home/valery/.local/share/gem/ruby/3.2.0/gems/turbo-rails-1.3.3/app/assets/javascripts
  /home/valery/.local/share/gem/ruby/3.2.0/gems/importmap-rails-1.1.5/app/assets/javascripts
  /home/valery/.rvm/gems/ruby-3.2.0/gems/actiontext-7.0.4.2/app/assets/javascripts
  /home/valery/.rvm/gems/ruby-3.2.0/gems/actiontext-7.0.4.2/app/assets/stylesheets
  /home/valery/.rvm/gems/ruby-3.2.0/gems/actioncable-7.0.4.2/app/assets/javascripts
  /home/valery/.rvm/gems/ruby-3.2.0/gems/activestorage-7.0.4.2/app/assets/javascripts
  /home/valery/.rvm/gems/ruby-3.2.0/gems/actionview-7.0.4.2/lib/assets/compiled
  /home/valery/projects/veda/tc-tech-platform/tc-tech-server/app/javascript
  /home/valery/projects/veda/tc-tech-platform/tc-tech-server/vendor/javascript
  /home/valery/.local/share/gem/ruby/3.2.0/gems/popper_js-2.11.6/assets/javascripts
  /home/valery/.local/share/gem/ruby/3.2.0/gems/bootstrap-5.2.3/assets/stylesheets
  /home/valery/.local/share/gem/ruby/3.2.0/gems/bootstrap-5.2.3/assets/javascripts
```

justinfrench:
[...]

PR #1393, opened 2025-07-14
Cleanup legacy CSS & Asset Pipeline issues

justinfrench:
Reviewing #1367 alerted me to a few "legacy CSS" issues we need to clean-up:

1. We're still shipping stylesheets for IE6 & IE7!
2. We're still talking about a Rails 3 era asset pipeline in the README
3. We're still shipping the CSS as a Rails Engine asset, which I don't *think* is compatible with the various modern asset options for Rails (other than perhaps sprockets)

So this PR:

* Adds a `formtastic:stylesheets` generator that copies `formtastic.css` into the app's assets
* Removes the Rails Engine, including `formtastic.css` and the two IE-specific sheets
* Simplifies the README to focus on the generator, rather than adding to the pipeline, conditional stylesheets, etc

This is a backwards incompatible change that 6.0 is the right time for, but it also might be a bit surprising to developers as part of the upgrade. I'm open to ideas on how to smooth this out, but:

* Pretty sure most people are only using Formtastic in the context of ActiveAdmin these days ;) 
* I can't think of a reasonable and non-noisy way to warn/deprecate this in the 5.0 branch
* Pretty sure most people would roll their own CSS — formtastic.css was a proof of concept
* It's pretty easy for apps to recover from (add the stylesheet(s) to your own app)


mkllnk:
This broke our app! :wink: 

Still using Sprockets, with Rails 8.1 and we still had the old `require formtastic` in our SCSS file. This line fixed it though:

```
./bin/rails generate formtastic:stylesheets
```

Easy enough when I read the Readme. It may be helpful to mention this in the release notes under highlights though. What do you think?

justinfrench:
Thanks @mkllnk, completely agree, I’ll add to release notes right now.

PR #1394, opened 2025-07-14
Add helper methods for semantic_errors building logic

TimeTravelerFromNow:
### Description
This refactor was identified in #1388
We wanted to clean up the logic in the `semantic_errors(args*)` within the method

### Solution
Refactor the semantic_errors helper which does the following: 
- First get the array from the model's base errors with the `semantic_error_list_from_base` method
- Second get the array of attribute errors such as `:title, :synopsis, :isbn` when using this helper 
`<%= f.semantic_errors :title, :synopsis, :isbn, :publish_year %>`. 

  - `semantic_error_list_from_attributes(args)` encapsulates this logic.
- Render the result within a `<ul>` with each error as a `<li>` (Unchanged)

### Testing
- Tests in rspec pass
- Tested validation errors manually in a form, and also without errors
Screenshot 2025-07-14 at 3 41 57 PM

### DIFF
--- a/lib/formtastic/helpers/errors_helper.rb
+++ b/lib/formtastic/helpers/errors_helper.rb
@@ -41,24 +41,19 @@ module Formtastic
       #   <% end %>
       def semantic_errors(*args)
         html_options = args.extract_options!
-        args = args - [:base]
-        full_errors = args.inject([]) do |array, method|
-          attribute = localized_string(method, method.to_sym, :label) || humanized_attribute_name(method)
-          errors = Array(@object.errors[method.to_sym]).to_sentence
-          errors.present? ? array << [attribute, errors].join(" ") : array ||= []
-        end
-        full_errors << @object.errors[:base]
-        full_errors.flatten!
-        full_errors.compact!
-        return nil if full_errors.blank?
         html_options[:class] ||= "errors"
+
+        full_errors = semantic_error_list_from_base
+        full_errors += semantic_error_list_from_attributes(args)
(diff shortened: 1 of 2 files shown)


TimeTravelerFromNow:
@justinfrench this is my first attempt at the refactor you suggested https://github.com/formtastic/formtastic/pull/1388#pullrequestreview-3013726346.
It reorganizes the existing logic into methods, any thoughts on this approach?
I added a comment above `Array(@object.errors[:base])`, since I did not want to change existing functionality with this PR however I wanted to flag that it's possible that `@object.errors[:base]` should always be an array in Rails

lib/formtastic/helpers/errors_helper.rb
```diff
@@ -77,6 +72,41 @@ def has_errors?(method, options)
       def render_inline_errors?
         @object && @object.respond_to?(:errors) && Formtastic::FormBuilder::INLINE_ERROR_TYPES.include?(inline_errors)
       end
+
+      def semantic_error_list_from_base
+        if @object.errors[:base].is_a?(Array)
+          @object.errors[:base]
+        else
```
justinfrench, line 79: There was a time where `errors` could return a string, but I can't find any evidence of that still being true, and wasn't able to reproduce in Rails console with Rails 8. Please do check my homework, but if you agree, then feel free to remove this!

TimeTravelerFromNow: @justinfrench alright, I committed this and updated the tests

Issue #1396, opened 2025-08-07
`semantic_errors` renders only base errors by default


TimeTravelerFromNow:
A nice feature would be for `semantic_errors` to render `:base` errors and all attribute errors by default.

Currently you have to pass something like `@object.errors.attribute_names`, or `*f.object.errors.keys` in order to render all errors

in our codebase, we're using 
`<%= f.error_messages %>`
from the deprecated form helper gem
https://github.com/joelmoss/dynamic_form

I'd like to change usage of `<%= f.error_messages %>` to the formtastic helper `<%= f.semantic_errors %>`. However I dont want to pass in `*f.object.errors.keys` every time so this default would be nice to have.

@justinfrench , do you have any opinions on this?

PR #1398, opened 2025-08-08
Change `semantic_errors` to render all errors by default

TimeTravelerFromNow:
Idea for solving #1396 

This will alter the default behavior of `semantic_errors`, instead `:base` being the default, if no args are passed, the `form.semantic_errors` helper will render all base and errored attributes.

I added some test objects without mocking, since I wanted to see an active model record's messaging by default.

Did some experimenting in a Rails 8 app console about `attribute_names`
```
 b = Book.new(synopsis: 'fake @@@')
=> 
#<Book:0x0000000111531558
...
formtastic-inline-errors(dev)> b.errors.attribute_names
=> []
formtastic-inline-errors(dev)> b.valid?
=> false
formtastic-inline-errors(dev)> b.errors.attribute_names
 b.errors.class
=> ActiveModel::Errors
formtastic-inline-errors(dev)> b.errors.attribute_names
=> [:title, :isbn, :base, :synopsis]
```

So I think the `attribute_names` should be safe to use, are there other types of classes passed to formtastic that wouldnt be ActiveModel::Errors?

### DIFF
--- a/lib/formtastic/form_builder.rb
+++ b/lib/formtastic/form_builder.rb
@@ -48,6 +48,7 @@ module Formtastic
     configure :priority_time_zones, []
     configure :required_string, proc { %{<abbr title="#{Formtastic::I18n.t(:required)}">*</abbr>}.html_safe }
     configure :semantic_errors_link_to_inputs, false
+    configure :semantic_errors_render_all_attributes, false
     configure :skipped_columns, [:created_at, :updated_at, :created_on, :updated_on, :lock_version, :version]
     configure :use_required_attribute, false
 
--- a/lib/formtastic/helpers/errors_helper.rb
+++ b/lib/formtastic/helpers/errors_helper.rb
@@ -9,7 +9,7 @@ module Formtastic
       INLINE_ERROR_TYPES = [:sentence, :list, :first]
 
       # Generates an unordered list of error messages on the base object and optionally for a given
-      # set of named attribute. This is idea for rendering a block of error messages at the top of
+      # set of named attributes. This is ideal for rendering a block of error messages at the top of
       # the form for hidden/special/virtual attributes (the Paperclip Rails plugin does this), or
       # errors on the base model.
       #
@@ -19,6 +19,8 @@ module Formtastic
       # # in config/initializers/formtastic.rb
       # Setting `Formtastic::FormBuilder.semantic_errors_link_to_inputs = true`
       # will render attribute errors as links to the corresponding errored inputs.
+      # Setting `Formtastic::FormBuilder.semantic_errors_render_all_attributes = true`
+      # will render base errors and all errored attributes when no arguments are passed
       #
       # @example A list of errors on the base model
       #   <%= semantic_form_for ... %>
@@ -43,9 +45,18 @@ module Formtastic
       #     <%= f.semantic_errors :something_special, :something_else, :class => "awesome", :onclick => "Awesome();" %>
       #     ...
       #   <% end %>
+      #
+      # @param [Array<Symbol>] *args Optional attribute names to display errors for.
+      #   When empty, displays base errors or full errors based on configuration. HTML options can be passed
+      #   as the last argument hash.
+      # @return [String, nil] HTML string containing error list, or nil if no errors exist
       def semantic_errors(*args)
         html_options = args.extract_options!
         html_options[:class] ||= "errors"
+
+        if Formtastic::FormBuilder.semantic_errors_render_all_attributes && args.empty?
+          args = @object.errors.attribute_names
+        end
 
         if Formtastic::FormBuilder.semantic_errors_link_to_inputs
           attribute_error_hash = semantic_error_hash_from_attributes(args)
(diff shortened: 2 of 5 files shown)


justinfrench:
@TimeTravelerFromNow delayed thanks for this work 🙏. You'll need to work through some conflicts against main (sorry), but I'm also wondering if a setting is the right approach. We could:

* change behaviour (I agree with @mikz this wasn't the right approach)
* introduce a configuration like you have here
* introduce a whole other method that does this (`semantic_errors_on_model`?)
* introduce a new special arg like `:base` (`:all`?)

I guess the thing that might help nail down the right API is to understand why you want this as a _default_? Is it a genuine preference (e.g.: _you always want all errors, and you don't want to have to type it each time_), or is it maybe because you _can't really_ type what you want (maybe there's something restrictive about ActiveAdmin or another system I don't understand, or something else?

TimeTravelerFromNow:
@justinfrench thanks for the review. No prob on the rebase, congrats on getting the latest versions out in the world!

> I guess the thing that might help nail down the right API is to understand why you want this as a default? Is it a genuine preference (e.g.: you always want all errors, and you don't want to have to type it each time)

On the question why I'd like this default, it's convenience. Just genuine preference: I'd always want all errors because I dont want to type them all out each time. When I see `<%= f.semantic_errors *f.object.errors.attribute_names %>` in forms this adds a small amount of cognitive load because a new reader of the code may want to investigate what this does, how, and why it needs all attribute names as an argument.

If a new reader saw `<%= f.semantic_errors :all %>`, that would probably be good enough to convey that this error summary handles all of the model's errors.

> is it maybe because you can't really type what you want (maybe there's something restrictive about ActiveAdmin or another system I don't understand, or something else?

Nope, all the scenarios I've worked in, I've always had access to all the object's attributes.

I like the approach of a special argument `:all`, but will clash in the rare case where a model has an errored attribute with the name `:all`.
If we're ok with that edge case, I'll proceed with implementing showing all errors when using it like so `<%= f.semantic_errors :all %>`. 

I wouldnt want to create new method, because it would be nearly all the same code as `semantic_errors`, but I'm ok with this approach if anyone insists.

@mikz what are your thoughts on `<%= f.semantic_errors :all %>` handling showing an error summary of all model errored attributes?

mikz:
On the topic of what is the default.

If we take example from a readme:
```
  <%= semantic_form_for @user do |f| %>
    <%= f.inputs :title, :body, :section, :categories, :created_at %>
    <%= f.actions :submit, :cancel %>
  <% end %>
  ```
  
I think you get inline errors for the title, body, section, categories and created_at. Then what I want from `f.semantic_errors` is to default to `:base`, not to duplicate other inline errors.

Thats where we diverge - you would like all forms to duplicate inline errors and have a list of all errors. 
I absolutely understand different design angles and that it even might be better UX, so that makes absolute sense.

But you probably have that per form, or set globally for the app. So then you might want that setting to be global or overridable on the form itself (via opts or form builder).

That is the difference right? Do you also want to show the inline errors when you're showing the list at the top for all of them? Would you consider disabling the inline errors when you show them all in one?

On the `:all` - it might not be the right abstraction here and a more global option would be a better choice - since that your design choice is maybe at a higher level.

TimeTravelerFromNow:
@mikz Thanks for the thoughtfulness on this.

> I think you get inline errors for the title, body, section, categories and created_at. Then what I want from f.semantic_errors is to default to :base, not to duplicate other inline errors.

Correct, I understand now why `:base` is a good default in the case that your UX and design achieves, which will avoid duplicating the inline errors both next to the errored inputs and the error summary.

> Thats where we diverge - you would like all forms to duplicate inline errors and have a list of all errors.
> I absolutely understand different design angles and that it even might be better UX, so that makes absolute sense.

Exactly, in my case, I would like all forms to duplicate inline errors and have a list of the errors in the summary. I just peeked into our codebase, and we're currently summarizing all errors at the top of the form in addition to inline errors. We're doing this in over 300 existing forms, by utilizing a deprecated form helper gem - _dynamic_form_. This gem provides us a form helper `<%= f.error_messages %>` which does almost the same thing as formtastic's semantic_errors helper, without the additional nice to have's within formtastic. I would like to migrate all 300 uses of `<%= f.error_messages %>` and replace them with `<%= f.semantic_errors %>`. Hopefully that helps you further understand where I'm at and the motivation behind this PR.

> Would you consider disabling the inline errors when you show them all in one?

Nope I havent considered disabling inline errors when showing the error summary with `semantic_errors`. If this becomes a need later, I could revisit, but there's no issue with duplicating them with our current design guidelines.

We're pretty closely following the UK gov's design patterns for our error UX https://design-system.service.gov.uk/components/error-summary/

Screenshot 2026-03-07 at 2 43 05 PM

> On the :all - it might not be the right abstraction here and a more global option would be a better choice - since that your design choice is maybe at a higher level.

Agreed, all of our 300 forms are using the same error summary design currently which shows a list of errors for all attributes at the top within the summary (in addition to `:base` errors), and the inline errors are also rendered next to inputs. Because we are consistently doing this across all of our forms, the global option makes the most sense to me.

TimeTravelerFromNow:
@mikz you convinced me that an opt-in config option is the right way forward for my app, and can ensure your UX + design system continues working the same way as well where the `:base` is default when no args are passed to `<%= f.semantic_errors %>`

Setting this once globally for my app will avoid any further abstractions that I need to add to formtastic and my html erb.
`Formtastic::FormBuilder.semantic_errors_render_all_attributes = true`

PR #1400, opened 2025-12-18
Add support for PostgreSQL timestamptz column type

Yarroo:
## Problem
After updating the pg gem, PostgreSQL introduces a new `timestamptz` column type.
Formtastic fails with "Unable to find input class TimestamptzInput" error.

## Solution
Map `:timestamptz` type to `:datetime_select` input (same as `:datetime` and `:timestamp`).

## Testing
- Added spec for timestamptz input mapping
- All existing tests pass
- Verified with ActiveAdmin forms using timestamptz columns


justinfrench:
Sorry for the delay, but this has been merged now. Thank you!

